PR #3, opened 2020-03-10
Add namespaces to avoid clashes with classes in consumer projects

kpdowns:
The specific use case for this PR is that when trying to use this client in my own project, the `Config` class clashes with another `Config` class specified in my project. Therefore, I've added in the same namespace as what's defined in the other classes so that it won't clash going forward.

### DIFF
--- a/Config.cs
+++ b/Config.cs
@@ -1,6 +1,9 @@
 using System;
 
-public static class Config {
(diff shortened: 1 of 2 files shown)


ssg-luke:
Looks good! Thanks for the contribution @kpdowns

kpdowns:
Thanks for merging this @ssg-luke !

PR #4, opened 2020-03-10
Allow API url to be defined when creating an instance of the BulletTrainClient

kpdowns:
This pull request is dependent on https://github.com/BulletTrainHQ/bullet-train-dotnet-client/pull/3 as the namespace changes are included when making the style consistent across all files (there was a variation in style where 1 file used tabs, the other spaces).

When constructing the client, it is not possible (from what I saw) to specify the BulletTrain API url. This is a necessity for me as I am currently trying to host BulletTrain locally. I'd thought that others might be having the same problem which could discourage people from trying this client library. I'd recommend that if this pull request is accepted, it be released as a v2 as it breaks existing consumers when constructing an instance of the client.

I was not sure of the exact purpose of the `ServicePointManager` in the context of this library, but I left the code as is and tried to maintain the functionality with regards to how you are creating a static instance of the `HttpClient`.

### DIFF
--- a/BulletTrainClient.cs
+++ b/BulletTrainClient.cs
@@ -23,6 +23,11 @@ namespace SolidStateGroup.BulletTrain
                     "Parameter must be provided when constructing an instance of the client.");
             }
 
+            if (!bulletTrainConfiguration.IsValid())
+            {
+                throw new ArgumentException("The provided configuration is not valid. An API Url and Environment Key must be provided.", nameof(bulletTrainConfiguration));
+            }
+
             if (!isInitialized)
             {
                 configuration = bulletTrainConfiguration;
--- a/BulletTrainClient.cs
+++ b/BulletTrainClient.cs
@@ -14,9 +14,10 @@ namespace SolidStateGroup.BulletTrain
 
         private static readonly HttpClient httpClient;
 
-        static BulletTrainClient() {
+        static BulletTrainClient()
+        {
             var sp = ServicePointManager.FindServicePoint(new Uri(Config.API));
-            sp.ConnectionLeaseTimeout = 60*1000*5;
+            sp.ConnectionLeaseTimeout = 60 * 1000 * 5;
             httpClient = new HttpClient();
         }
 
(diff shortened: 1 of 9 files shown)


dabeeeenster:
Thanks for the 2 pull requests - much appreciated. We'll take a look and hopefully can merge a 2.0 soon 👍

kpdowns:
Thanks @dabeeeenster. Happy to make any changes required 👍

BulletTrainClient.cs
```diff
@@ -10,14 +10,27 @@ namespace SolidStateGroup.BulletTrain
...
+        private static HttpClient httpClient;
+        private static bool isInitialized;
 
-        static BulletTrainClient() {
-            var sp = ServicePointManager.FindServicePoint(new Uri(Config.API));
-            sp.ConnectionLeaseTimeout = 60*1000*5;
-            httpClient = new HttpClient();
+        public BulletTrainClient(BulletTrainConfiguration bulletTrainConfiguration)
+        {
+            if (bulletTrainConfiguration == null)
```
ssg-luke, line 20: @kpdowns It'd be great here if after this null check you could check and throw errors for when the EnvironmentKey and ApiUrl is null or empty

kpdowns: Hi @ssg-luke . Happy to make that change. What would you prefer? We could encapsulate the validation into the `BulletTrainConfiguration` object? I.e.

```
    public class BulletTrainConfiguration
    {
        public string ApiUrl = "https://api.bullet-train.io/api/v1/";
        public string EnvironmentKey = "";

        public void Validate() 
        {
            if(string.IsNullOrEmpty(ApiUrl))
            .....
        }
    }
```

Or simply doing it inline after that initial null check?

Going down the 1st path would keep the logic for validating the `BulletTrainConfiguration` coupled with an instance of the `BulletTrainConfiguration`. If the object then happens to become larger in future, it would be possible to add more checks in only one place, and it would technically allow you to validate the configuration before you try to create an instance of the client.

ssg-luke: 1st path sounds good alongside calling that Validate() func after the initial null check in the BulletTrainClient constructor. I believe if Validate() then throws an exception it should also cause failure to construct the client without needing to necessarily ever call the Validate() func on the config object or alternatively have Validate() return bool and throw a generic exception for bad config in the constructor

kpdowns: Ok great, I'm inclined to go for having `Validate` return a bool. Then after the null check when creating the client, call that. If it's not valid, then throw an exception during construction of the client. I'll make that change shortly.

kpdowns: Updated, let me know if you'd like any other changes.

ssg-luke: Nope its looking good! Will get a v2.0 out when I find some time

kpdowns: Thanks @ssg-luke !

PR #8, opened 2020-10-28
fix: problems with configuration of API

Menur:
The issue is $"{configuration.ApiUrl}flags/" can end up in the wrong URL when `ApiUrl` doesn't end with `/`.

It's the same as with the path, it should be better to combine it properly.


dabeeeenster:
Thanks!

PR #13, opened 2021-07-11
fix: return empty string instead of null when http request fails

xenm:
this avoids unhandled ArgumentNullExceptions coming from Newtonsoft.JsonSerializer

### DIFF
--- a/BulletTrainClient.cs
+++ b/BulletTrainClient.cs
@@ -295,7 +295,7 @@ namespace BulletTrain
             {
                 Console.WriteLine("\nHTTP Request Exception Caught!");
                 Console.WriteLine("Message :{0} ", e.Message);
-                return null;
+                return string.Empty;
             }
         }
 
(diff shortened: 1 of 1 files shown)


BulletTrainClient.cs
```diff
@@ -295,7 +295,7 @@ private async Task<string> GetJSON(HttpMethod method, string url, string body =
             {
                 Console.WriteLine("\nHTTP Request Exception Caught!");
                 Console.WriteLine("Message :{0} ", e.Message);
-                return null;
+                return string.Empty;
```
Menur, line 298: Check also[ this line](https://github.com/Flagsmith/flagsmith-dotnet-client/blob/d3b43b5378fdaf7fdcf4552983e029b850577023/BulletTrainClient.cs#L125) as `string.Empty`is deserialized as null.

xenm: Thanks, didn't see that. Fixed.

dabeeeenster:
Thanks so much!

PR #17, opened 2021-07-21
feat: return null instead of default value for value types

xenm:
This way anyone using this client can distinguish if BT was not responding or if the FF value was false / zero
Changing version to 2.2.0 as return types were updated

### DIFF
--- a/BulletTrainClient.cs
+++ b/BulletTrainClient.cs
@@ -81,12 +81,13 @@ namespace BulletTrain
         /// <summary>
         /// Check feature exists and is enabled optionally for a specific identity
         /// </summary>
-        public async Task<bool> HasFeatureFlag(string featureId, string identity = null)
+        /// <returns>Null if Bullet Train is unaccessible</returns>
+        public async Task<bool?> HasFeatureFlag(string featureId, string identity = null)
         {
             List<Flag> flags = await GetFeatureFlags(identity);
             if (flags == null)
             {
-                return false;
+                return null;
             }
 
             foreach (Flag flag in flags)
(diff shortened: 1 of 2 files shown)


dabeeeenster:
Hi,

Thanks for this. We're not .Net experts, but what be more idiomatic to throw an exception if there was no reply from the API?

xenm:
Hi,

It is true that this is discutable.

Exception is always valid output. But often not the best practice. Avoiding of throwing exceptions is a bit quicker and whoever is using this client don't have to catch them manually in every use. Hence moving some of the logic back inside the client lets say. Another practice is to throw exceptions only in extreme situations. Not when service is temporarily unavailable and users will want to act accordingly.

The original problem was that from this client were coming unhandled parsing exceptions. And these exceptions were not mentioned anywhere so users wouldn't expect them and handle them properly.

So at this point we are returning null when it makes sense. (GetCollection returns null or the collection of required entities) And to implement default values for Feature Flags at user's end - this is all we need as a response.

Now the problem is with public methods returning value of a specific FF. We need 1 more value to propagate that the FF's value is unknown.

It is completely fine (but slower) to solve this with proper exception. (FlagsmithUnavailableException rather than NullArgumentException for example) But now some methods will return null and others will throw an exception. Which brings more confusion to the users.


So at this point I think it is more convenient to make this situation exception free. Rather than throwing exceptions in every method. What do you think?

Another thing to consider here is logging of these errors/situations. (And decision on way of logging can impact the above ideas) Maybe allow DI of loggers instead of writing to console.

PR #23, opened 2022-02-15
Feature/rewrite for client side eval

asadabbasfolio3:
- Integration for Engine with Client
- Written Test Cases for Client-Engine Integration
- naming tweaks updation
- updated workflow to run on all specified Frameworks.


Flagsmith.FlagsmithClient/AnalyticFlag.cs
```diff
@@ -0,0 +1,45 @@
+﻿using FlagsmithEngine.Feature.Models;
+using System.Collections.Generic;
+using System.Linq;
+using System;
+using Newtonsoft.Json;
+using System.Threading.Tasks;
+
+namespace Flagsmith
+{
+    public class AnalyticFlag : Flag
```
matthewelwell, line 10: I'm not sure why we need this separate class? Couldn't this logic be contained within the regular `Flag` class?

Flagsmith.FlagsmithClient/Extensions/DictionaryExtensions.cs
```diff
@@ -0,0 +1,17 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Flagsmith.Extensions
+{
+    internal static class DictionaryExtensions
```
matthewelwell, line 7: What's the purpose of this class? Is this some C# magic to override the `ForEach` method on a dictionary?

Flagsmith.FlagsmithClient/Feature.cs
```diff
@@ -0,0 +1,30 @@
+using Newtonsoft.Json;
+
+namespace Flagsmith
+{
+    [JsonObject(MemberSerialization.OptIn)]
+    public class Feature
```
matthewelwell, line 6: I'm not too sure why we need this class in the Client project? The python client just uses the feature id and feature name rather than needing a whole class to hold the information.

Flagsmith.FlagsmithClient/Flag.cs
```diff
@@ -0,0 +1,52 @@
...
+        public Flag(Feature feature, bool enabled, string value)
+        {
+            this.Feature = feature;
+            this.Enabled = enabled;
+            this.Value = value;
+        }
+        [JsonProperty("featureId")]
+        protected int FeatureId;
+        [JsonProperty("feature")]
+        protected Feature Feature = null;
```
matthewelwell, line 23: As per above comment, this should just be `protected string FeatureName;` I think?

Flagsmith.FlagsmithClient/FlagsmithClient.cs
```diff
@@ -0,0 +1,437 @@
...
+    /// </summary>
+    /// <exception cref="FlagsmithAPIError">
+    /// Thrown when error occurs during any http request to Flagsmith api.Not applicable for polling or ananlytics.
+    /// </exception>
+    /// <exception cref="FlagsmithClientError">
+    /// A general exception with a error message. Example: Feature not found, etc.
+    /// </exception>
+    public class FlagsmithClient
+    {
+        public static FlagsmithClient instance;
```
matthewelwell, line 31: So the client has to maintain an instance of itself? 🤔 What's the reason for this?

Flagsmith.FlagsmithClient/FlagsmithClient.cs
```diff
@@ -0,0 +1,437 @@
...
+            }
+
+            if (instance == null)
+            {
+                configuration = flagsmithConfiguration;
+                var sp = ServicePointManager.FindServicePoint(new Uri(configuration.ApiUrl));
+                sp.ConnectionLeaseTimeout = 60 * 1000 * 5;
+                httpClient = new HttpClient();
+                instance = this;
+                _PollingManager = new PollingManager(GetAndUpdateEnvironmentFromApi, configuration.EnvironmentRefreshIntervalSeconds);
```
matthewelwell, line 60: Shouldn't we move this within the `if (configuration.EnableClientSideEvaluation)` conditional block?

Flagsmith.FlagsmithClient/FlagsmithClient.cs
```diff
@@ -0,0 +1,437 @@
...
+                _Engine = new Engine();
+                if (configuration.EnableAnalytics)
+                    _AnalyticsProcessor = new AnalyticsProcessor(httpClient, configuration.EnvironmentKey, configuration.ApiUrl, configuration.Logger, configuration.CustomHeaders);
+                if (configuration.EnableClientSideEvaluation)
+                    _ = _PollingManager.StartPoll();
+
+            }
+            else
+            {
+                throw new NotSupportedException("FlagsmithClient should only be initialised once. Use FlagsmithClient.instance after successful initialisation");
```
matthewelwell, line 70: What's the reason for this?

Flagsmith.FlagsmithClient/FlagsmithClient.cs
```diff
@@ -0,0 +1,437 @@
...
+            }
+        }
+
+        /// <summary>
+        /// Get all the default for flags for the current environment.
+        /// </summary>
+        public async Task<List<Flag>> GetFeatureFlags()
```
matthewelwell, line 77: As per all our discussions, this class should only have 2 public methods (+ the one to update environment). Those methods should be named identically (barring snake / pascal case differences) to the ones in the python client. 

Please can you rewrite this client class so that it has only the methods available in the python client? Those methods should, as per the python client, return an object, from which you can access methods such as `IsFeatureEnabled` and `GetFeatureValue`. Those methods should not live on the client itself.

Flagsmith.FlagsmithClient/FlagsmithConfiguration.cs
```diff
@@ -0,0 +1,61 @@
+﻿using Microsoft.Extensions.Logging;
+using System;
+using System.Collections.Generic;
+namespace Flagsmith
+{
+    public class FlagsmithConfiguration
+    {
+        public FlagsmithConfiguration()
+        {
+            ApiUrl = "https://api.flagsmith.com/api/v1/";
```
matthewelwell, line 10: Does this set the default for all constructors or only if you use the 'no args constructor' that is defined here? Can we not set the default on the attribute definition below?

Flagsmith.FlagsmithClient/FlagsmithConfiguration.cs
```diff
@@ -0,0 +1,61 @@
...
+            EnvironmentKey = string.Empty;
+            UseLegacyIdentities = true;
```
matthewelwell, line 12: What's this?

matthewelwell (changes requested):
@asadabbasfolio3 I've been through all of the non-test files for now and added a few comments. The key thing is that the implementation of the Client needs quite a lot of changes. The Client implementation you have seems closer to the old .NET client, however, we would like it to be simply a reimplementation of the python client. Please can you have another look at the implementation in python and update the client accordingly. Feel free to give me a me a shout on Slack if you want me to go through it with you. 

Also, I think we now have 2 example apps, can you have a look and make sure we only have one please?

Example/Model/Search.cs
```diff
@@ -0,0 +1,9 @@
+﻿namespace Example.Model
+{
+    public class search
```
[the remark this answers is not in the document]
asadabbasfolio3: we are using this class in Example for retrieving body params. should be pascal,i will update that.

Example/Extensions/FlagsmithExtensions.cs
```diff
@@ -0,0 +1,25 @@
...
+{
+    public static class FlagsmithExtensions
+    {
+        public static void RegisterFlagsmithClientAsSingleton(this IServiceCollection services, IConfiguration configuration)
+        {
+            var settings = configuration.GetSection("FlagsmithConfiguration").Get<FlagsmithSettings>();
+            var flagsmithClient = new FlagsmithClient(MapFlagsmithConfiguration(settings));
+            services.AddSingleton<FlagsmithClient>(flagsmithClient);
+        }
+        private static FlagsmithConfiguration MapFlagsmithConfiguration(FlagsmithSettings settings)
```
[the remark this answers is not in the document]
asadabbasfolio3: its conventional way to putting your configurations in `appsettings.json` , we are retrieving it from here and creating a singleton of `FlagsmithClient` object , it later i can easily we used by simple Dependency Injection(DI).

Flagsmith.FlagsmithClient/FlagsmithClient.cs
asadabbasfolio3: yes we should, will update that.

Flagsmith.FlagsmithClient/FlagsmithClient.cs
asadabbasfolio3: previously in dotnet , we were insuring that client instance should be created once, shouldn't it be singleton or allow to create multiple ?

Flagsmith.FlagsmithClient/FlagsmithConfiguration.cs
asadabbasfolio3: we only have default constructor ,which just initialized default values of properties.we can set on attribute level but conventionally dotnet recommend it to be initialized at constructor level.

Flagsmith.FlagsmithClient/FlagsmithConfiguration.cs
asadabbasfolio3: for `UseLegacyIdentities`, I missed to discuss that property, as in previous dotnet implementation this property is used to call either depreciated or new Identities API url in `GetIdentitiesUrl()`. should i remove that ?

Example/Program.cs
```diff
@@ -0,0 +1,36 @@
+using Example.Extensions;
+using Flagsmith;
+var builder = WebApplication.CreateBuilder(args);
+builder.Services.RegisterFlagsmithClientAsSingleton(builder.Configuration);
+var app = builder.Build();
+
+app.MapGet("/", async (FlagsmithClient flagsmithClient) =>
```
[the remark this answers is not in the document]
asadabbasfolio3: Sure i will update that example like we did in python.

.github/workflows/formatting-and-tests.yml
```diff
@@ -36,19 +42,28 @@ jobs:
         strategy:
             max-parallel: 4
             matrix:
-                dotnet-version: ["3.1", "5.0", "6.0"]
-                working-directory: ["Flagsmith.DotnetClient", "Flagsmith.Engine","Flagsmith.EngineTest"]
+                working-directory: ["Flagsmith.EngineTest","Flagsmith.Clent.Test"]
```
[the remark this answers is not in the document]
asadabbasfolio3: will correct that.

.github/workflows/formatting-and-tests.yml
```diff
@@ -7,12 +7,12 @@ on:
...
     formatting:
         runs-on: ubuntu-latest
-        name: Check formatting
+        name: Check Build and formatting
 
         strategy:
           max-parallel: 2
           matrix:
-            working-directory: ["Flagsmith.DotnetClient", "Flagsmith.Engine","Flagsmith.EngineTest"]
+            working-directory: ["Flagsmith.FlagsmithClient", "Flagsmith.Engine","Flagsmith.EngineTest","Flagsmith.Clent.Test"]
```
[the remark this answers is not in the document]
asadabbasfolio3: will correct that.

Example/appsettings.json
```diff
@@ -0,0 +1,19 @@
+{
+  "Logging": {
+    "LogLevel": {
+      "Default": "Information",
+      "Microsoft.AspNetCore": "Warning"
+    }
+  },
+  "FlagsmithConfiguration": {
+    //"EnvironmentKey": "kxsJgKrjzFhxzjdL6bmQMp",
+    "EnvironmentKey": "ser.ngVTUgJjqEWc3qKkj7W4Qj",
```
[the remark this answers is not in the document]
asadabbasfolio3: Yes this is the key i used to test , i should update it to placeholder.

Flagsmith.FlagsmithClient/Extensions/DictionaryExtensions.cs
asadabbasfolio3: Dictionary Collection in c# dont have Foreach method , that's why we created extension method on our own.

Flagsmith.FlagsmithClient/FlagsmithClient.cs
asadabbasfolio3: just to ensure singleton , as i discussed here : [](https://github.com/Flagsmith/flagsmith-dotnet-client/pull/23#discussion_r811630716)

Flagsmith.FlagsmithClient/Feature.cs
asadabbasfolio3: as per API JSON response , we have an object of feature , so i find it convenient to create its separate class.

Flagsmith.FlagsmithClient/Flag.cs
asadabbasfolio3: should i replace it with Feature object ? because if we update it then JSON response and our Class schema will be different.

Flagsmith.FlagsmithClient/FlagsmithClient.cs
matthewelwell: Unless its standard behaviour in .NET I suggest that we don't worry about creating a singleton like this.

Flagsmith.FlagsmithClient/FlagsmithConfiguration.cs
matthewelwell: But, ok, what happens if the user does this: 

```dotnet
var configuration = FlagsmithConfiguration(EnvironmentKey="my-key")
```

Will it pick up the default API url?

Flagsmith.FlagsmithClient/FlagsmithConfiguration.cs
matthewelwell: Yes, this client should only use `/api/v1/identities/`.

Flagsmith.FlagsmithClient/FlagsmithConfiguration.cs
asadabbasfolio3: No, as its no parameter constructor , user won't be allowed to pass his own params inside.

Flagsmith.FlagsmithClient/FlagsmithClient.cs
asadabbasfolio3: Sure , will update it.

Flagsmith.FlagsmithClient/FlagsmithConfiguration.cs
matthewelwell: Then this solution really doesn't work. Most of our users will not want to care about specifying the ApiUrl but they obviously will need to specify the environment key. How can we make it such that this is possible?

asadabbasfolio3: I am removing `FlagsmithConfiguration.cs` and taking params directly from `flagsmithClient` constructor as we did in python implementation

matthewelwell: Ah, right, sorry! I forgot about the properties json approach in .NET. If that is a standard way to accomplish the initialisation here then I'm happy to stick with that approach.

matthewelwell:
Superseded by #25.

Flagsmith.FlagsmithClient/AnalyticFlag.cs
asadabbasfolio3: removed analyticFlag class , and updated its logic like python implementation.

PR #25, opened 2022-02-22
Feature/engine to client integration

asadabbasfolio3:
- Integration for Engine with Client
- Written Test Cases for Client-Engine Integration
- naming tweaks updation
- some suggested changes from [rewrite for client side eval](https://github.com/Flagsmith/flagsmith-dotnet-client/pull/23)


Example/Program.cs
```diff
@@ -0,0 +1,36 @@
+using Example.Extensions;
+using Flagsmith;
+var builder = WebApplication.CreateBuilder(args);
+builder.Services.RegisterFlagsmithClientAsSingleton(builder.Configuration);
+var app = builder.Build();
+
+app.MapGet("/", async (FlagsmithClient flagsmithClient) =>
```
matthewelwell, line 7: As discussed on the other PR, let's bring this closer to the python example so it looks more like a real world app. There should be a readme in this example directory too about how to run the example application.

Example/Settings/FlagsmithSettings.cs
```diff
@@ -0,0 +1,13 @@
+﻿namespace Example.Settings
+{
+    public class FlagsmithSettings
```
matthewelwell, line 3: I'm not too sure why we need a separate Settings class in the Example project. Shouldn't this just be part of the main project?

Flagsmith.Client.Test/AnalyticsProcessorTest.cs
```diff
@@ -0,0 +1,49 @@
...
+using System.Linq;
+using System.Net.Http;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Flagsmith.FlagsmithClientTest
+{
+    internal class AnalyticsProcessorTest : AnalyticsProcessor
+    {
+        Dictionary<string, int> _totalFucntionCalls;
```
matthewelwell, line 13:
```suggestion
        Dictionary<string, int> _totalFunctionCalls;
```

Flagsmith.Client.Test/AnalyticsProcessorTest.cs
```diff
@@ -0,0 +1,49 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Net.Http;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Flagsmith.FlagsmithClientTest
+{
+    internal class AnalyticsProcessorTest : AnalyticsProcessor
```
matthewelwell, line 11: I think this class is essentially subclassing the AnalyticsProcessor to facilitate testing, is that correct?

Is there a way that we can test the AnalyticsProcessor class without the need for this additional class? Although we aren't really changing the behaviour of the class itself, it certainly feels unusual to test our AnalyticsProcessor class indirectly through this class instead.

Flagsmith.Client.Test/FlagsmithClientTest.cs
```diff
@@ -0,0 +1,84 @@
...
+using System.Threading.Tasks;
+using Moq;
+using System.Net.Http;
+using System.Threading;
+using Newtonsoft.Json;
+using System.Linq;
+
+namespace Flagsmith.FlagsmithClientTest
+{
+    internal class FlagsmithClientTest : FlagsmithClient
```
matthewelwell, line 13: As above, do we _need_ this class?

Flagsmith.FlagsmithClient/AnalyticFlag.cs
```diff
@@ -0,0 +1,45 @@
+﻿using FlagsmithEngine.Feature.Models;
+using System.Collections.Generic;
+using System.Linq;
+using System;
+using Newtonsoft.Json;
+using System.Threading.Tasks;
+
+namespace Flagsmith
+{
+    public class AnalyticFlag : Flag
```
matthewelwell, line 10: Is there a reason that there is this additional class here? The python client handles the analytics processing on the `Flags` class which I don't think exists in your .NET implementation. Is there a specific reason for this?

Flagsmith.FlagsmithClient/Flag.cs
```diff
@@ -0,0 +1,52 @@
...
+        public Flag(Feature feature, bool enabled, string value)
+        {
+            this.Feature = feature;
+            this.Enabled = enabled;
+            this.Value = value;
+        }
+        [JsonProperty("featureId")]
+        protected int FeatureId;
+        [JsonProperty("feature")]
+        protected Feature Feature = null;
```
matthewelwell, line 23: As per the discussion on your previous PR, we should simplify this to just the string `featureName`. I'd like to keep the `Flag` class as slim as possible.

Flagsmith.FlagsmithClient/FlagsmithClient.cs
```diff
@@ -0,0 +1,431 @@
...
+    /// </summary>
+    /// <exception cref="FlagsmithAPIError">
+    /// Thrown when error occurs during any http request to Flagsmith api.Not applicable for polling or ananlytics.
+    /// </exception>
+    /// <exception cref="FlagsmithClientError">
+    /// A general exception with a error message. Example: Feature not found, etc.
+    /// </exception>
+    public class FlagsmithClient
+    {
+        public static FlagsmithClient instance;
```
matthewelwell, line 31: As per the discussion on the previous PR, let's remove this singleton logic.

Flagsmith.FlagsmithClient/FlagsmithClient.cs
```diff
@@ -0,0 +1,431 @@
...
+    /// A Flagsmith client.
+    /// Provides an interface for interacting with the Flagsmith http API.
+    /// </summary>
+    /// <exception cref="FlagsmithAPIError">
+    /// Thrown when error occurs during any http request to Flagsmith api.Not applicable for polling or ananlytics.
+    /// </exception>
+    /// <exception cref="FlagsmithClientError">
+    /// A general exception with a error message. Example: Feature not found, etc.
+    /// </exception>
+    public class FlagsmithClient
```
matthewelwell, line 29: As discussed on the other PR, this class should be trimmed right down to only expose the methods available on the python client.

Flagsmith.Client.Test/PollingManagerTest.cs
```diff
@@ -0,0 +1,30 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+using Xunit;
+using Moq;
+using System.Threading.Tasks;
+
+namespace Flagsmith.FlagsmithClientTest
+{
+    public class PollingManagerTest
```
matthewelwell, line 10: This isn't really directly testing the `PollingManager` is it? It's more testing the `FlagsmithClient` and _indirectly_ testing the `PollingManager`. Since the `PollingManager` class simply expects a callback function, is there not a way that we can pass some kind of mock function in to test it in isolation?

matthewelwell (commented):
Added some more comments and some reminders about conversations on the previous PR.

Example/Settings/FlagsmithSettings.cs
asadabbasfolio3: `FlagsmithSettings ` is used to get and set flagsmith properties from app configuration only.

Flagsmith.Client.Test/FlagsmithClientTest.cs
asadabbasfolio3: we can not test private methods directly in dotnet, that is why we keep private method as protected and calling the same function from their derived class.

matthewelwell: I'm not sure why we need to test any private methods though? I don't think the python client tests any private methods does it?

Flagsmith.Client.Test/AnalyticsProcessorTest.cs
asadabbasfolio3: we are actually not overriding the actual method to test in derived class , but for `analyticsData `as because our `analyticsData `dictionary is set to protected.

Flagsmith.FlagsmithClient/AnalyticFlag.cs
asadabbasfolio3: I have as per python implementation.

Flagsmith.FlagsmithClient/FlagsmithClient.cs
asadabbasfolio3: i have removed the singleton pattern from class.

Flagsmith.FlagsmithClient/FlagsmithClient.cs
asadabbasfolio3: updated the access modifiers in `flagsmithclient `, and set methods as public only which are needed.

Flagsmith.Client.Test/PollingManagerTest.cs
asadabbasfolio3: Now i have updated the code to directly test the `pollingManager`.

Flagsmith.FlagsmithClient/Flag.cs
asadabbasfolio3: working on it , as i would need to write custom serialization for that.

Example/Program.cs
matthewelwell: It's definitely closer, yes, but the python example uses an HTML template [here](https://github.com/Flagsmith/flagsmith-python-client/blob/release/3.0.0/example/templates/home.html).Is there an easy way to include a template like that in the dotnet example app?

Flagsmith.Client.Test/FlagsmithTest.cs
```diff
@@ -0,0 +1,186 @@
...
+        [Fact]
+        public void TestFlagsmithStartsPollingManagerOnInitIfEnabled()
+        {
+            var mockHttpClient = HttpMocker.MockHttpResponse(new HttpResponseMessage
+            {
+                StatusCode = System.Net.HttpStatusCode.OK,
+                Content = new StringContent(Fixtures.JsonObject.ToString())
+            });
+            var flagsmithClientTest = new FlagsmithClient(Fixtures.ApiKey, httpClient: mockHttpClient.Object, enableClientSideEvaluation: true);
+            mockHttpClient.Verify(x => x.SendAsync(It.IsAny<HttpRequestMessage>(), It.IsAny<CancellationToken>()), Times.Once);
```
matthewelwell, line 19: Can we use a mocking library here to mock the polling manager instead of checking that an HTTP request is made? 

If not, can we at least check which endpoint is being hit here? It's a bit confusing here as the name of the test is talking about starting the polling manager but then all we're checking is that an HTTP request is made 😕

Flagsmith.Client.Test/FlagsmithTest.cs
```diff
@@ -0,0 +1,186 @@
...
+        public async Task TestGetIdentityFlagsUsesLocalEnvironmentWhenAvailable()
+        {
+            var mockHttpClient = HttpMocker.MockHttpResponse(new HttpResponseMessage
+            {
+                StatusCode = System.Net.HttpStatusCode.OK,
+                Content = new StringContent(Fixtures.JsonObject.ToString())
+            });
+            var flagsmithClientTest = new FlagsmithClient(Fixtures.ApiKey, enableClientSideEvaluation: true, httpClient: mockHttpClient.Object);
+            mockHttpClient.Verify(x => x.SendAsync(It.IsAny<HttpRequestMessage>(), It.IsAny<CancellationToken>()), Times.Once);
+            _ = await flagsmithClientTest.GetIdentityFlags("identifier", null);
```
matthewelwell, line 89: There are no assertions here, I don't understand what we're testing?

Flagsmith.Client.Test/FlagsmithTest.cs
```diff
@@ -0,0 +1,186 @@
...
+        public async Task TestGetEnvironmentFlagsCallsApiWhenNoLocalEnvironment()
+        {
+            var mockHttpClient = HttpMocker.MockHttpResponse(new HttpResponseMessage
+            {
+                StatusCode = System.Net.HttpStatusCode.OK,
+                Content = new StringContent(Fixtures.ApiFlagResponse)
+            });
+            var flagsmithClientTest = new FlagsmithClient(Fixtures.ApiKey, httpClient: mockHttpClient.Object);
+            var flags = (await flagsmithClientTest.GetEnvironmentFlags()).AllFlags();
+            mockHttpClient.Verify(x => x.SendAsync(It.IsAny<HttpRequestMessage>(), It.IsAny<CancellationToken>()), Times.Once);
```
matthewelwell, line 44: Again, can we verify what network request was made here?

Flagsmith.Client.Test/FlagsmithTest.cs
```diff
@@ -0,0 +1,186 @@
...
+                StatusCode = System.Net.HttpStatusCode.OK,
+                Content = new StringContent(Fixtures.JsonObject.ToString())
+            });
+            var flagsmithClientTest = new FlagsmithClient(Fixtures.ApiKey, enableClientSideEvaluation: true, httpClient: mockHttpClient.Object);
+            mockHttpClient.Verify(x => x.SendAsync(It.IsAny<HttpRequestMessage>(), It.IsAny<CancellationToken>()), Times.Once);
+            var flags = (await flagsmithClientTest.GetEnvironmentFlags()).AllFlags();
+            var fs = Fixtures.Environment.FeatureStates[0];
+            Assert.Equal(fs.Enabled, flags[0].IsEnabled());
+            Assert.Equal(fs.GetValue(), flags[0].GetValue());
+            Assert.Equal(fs.Feature.Name, flags[0].GetFeature().GetName());
```
matthewelwell, line 63: Can we verify that a call was not made to `/api/v1/flags/` ?

Flagsmith.Client.Test/FlagsmithTest.cs
```diff
@@ -0,0 +1,186 @@
...
+            var mockHttpClient = HttpMocker.MockHttpResponse(new HttpResponseMessage
+            {
+                StatusCode = System.Net.HttpStatusCode.OK,
+                Content = new StringContent(Fixtures.ApiIdentityResponse)
+            });
+            var flagsmithClientTest = new FlagsmithClient(Fixtures.ApiKey, httpClient: mockHttpClient.Object);
+            var flags = (await flagsmithClientTest.GetIdentityFlags("identifier")).AllFlags();
+            Assert.True(flags[0].IsEnabled());
+            Assert.Equal("some-value", flags[0].GetValue());
+            Assert.Equal("some_feature", flags[0].GetFeature().GetName());
```
matthewelwell, line 77: Can we verify that a call was made to `/api/v1/identities/`?

Flagsmith.FlagsmithClient/FlagsmithClient.cs
matthewelwell: I'm confused why we still have `GetTraits`, `GetTrait`, methods etc. ?

Flagsmith.Client.Test/AnalyticsProcessorTest.cs
asadabbasfolio3: alright, i will write document to this class.

Flagsmith.FlagsmithClient/FlagsmithClient.cs
asadabbasfolio3: i thought to remain existing function , will be removing them.

Flagsmith.Client.Test/AnalyticsProcessorTest.cs
```diff
@@ -0,0 +1,27 @@
...
+    {
+        public bool IsFlushEarlyReturn { get; private set; } = false;
```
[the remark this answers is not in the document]
asadabbasfolio3: yes , removed it.

Flagsmith.Client.Test/FlagsmithTest.cs
```diff
@@ -0,0 +1,193 @@
...
+            var ss = Times.Once;
+            var mockHttpClient = HttpMocker.MockHttpResponse(new HttpResponseMessage
+            {
+                StatusCode = System.Net.HttpStatusCode.OK,
+                Content = new StringContent(Fixtures.JsonObject.ToString())
+            });
+            var flagsmithClientTest = new FlagsmithClient(Fixtures.ApiKey, enableClientSideEvaluation: true, httpClient: mockHttpClient.Object);
+            mockHttpClient.verifyHttpRequest(HttpMethod.Get, "/api/v1/environment-document/", Times.Once);
+            await flagsmithClientTest.GetEnvironmentFlags();
+            mockHttpClient.verifyHttpRequest(HttpMethod.Get, "/api/v1/environment-document/", Times.Once);
```
matthewelwell, line 33: I think perhaps the following would be more useful here? Perhaps we can add both? 

```dotnet
mockHttpClient.verifyHttpRequest(HttpMethod.Get, "/api/v1/flags/", Times.Never);
```

Flagsmith.FlagsmithClient/Flag.cs
```diff
@@ -0,0 +1,86 @@
...
+        public string Value { get; private set; }
+
+        public override string ToString()
+        {
+            return JsonConvert.SerializeObject(this);
+        }
+    }
+    internal class FlagJsonConverter : JsonConverter
+    {
+        public override object ReadJson(JsonReader reader, Type objectType, object existingValue, JsonSerializer serializer)
```
matthewelwell, line 47: Can we add some documentation to explain what this is doing?

Example/Settings/FlagsmithSettings.cs
matthewelwell: Do we still need this? Certainly it shouldn't still reference `UserLegacyIdentities`, right?

Flagsmith.FlagsmithClient/Flag.cs
matthewelwell: Can you elaborate a bit more? This method is not very readable, perhaps because I don't know .NET very well. I can't follow what it's doing and I _think_ I know what it's doing (somehow it's converting a returned API flag to a Flag object and flattening the `feature` attribute to set the `featureName` attribute?). I'm surprised that something that simple takes this much code but if it definitely does, I'd like to try and document it so that future developers can understand it's purpose. Perhaps some in line comments would help here?

Example/Settings/FlagsmithSettings.cs
asadabbasfolio3: yes just used example to model properties from `appsettings.json` .

`var settings = configuration.GetSection("FlagsmithConfiguration").Get<FlagsmithSettings>();`

Flagsmith.Client.Test/FlagsmithTest.cs
asadabbasfolio3: added

Flagsmith.FlagsmithClient/Flag.cs
asadabbasfolio3: added inline comments along with a reference link.

matthewelwell: As discussed in the standup this morning, let's return to having the separate `Feature` class and add some `getFeatureId()` and `getFeatureName()` methods to the `Flag` class.

asadabbasfolio3: as discussed removed converter and updated necessities

Flagsmith.FlagsmithClient/Flag.cs
```diff
@@ -0,0 +1,34 @@
...
+        public Flag(Feature feature, bool enabled, string value, int featureId = default)
+        {
+            this.Enabled = enabled;
+            this.Value = value;
+            this.Feature = feature;
+        }
+        [JsonProperty("id")]
+        public int Id { get; private set; }
+        [JsonProperty("feature")]
+        public Feature Feature { get; private set; }
```
matthewelwell, line 21: Can we add the getter methods for featureId and featureName as we discussed?

matthewelwell (approved):

PR #26, opened 2022-04-05
Allow null values for traits

matthewelwell:


FlagsmithClient.cs
```diff
@@ -237,9 +237,9 @@ public async Task<Trait> SetTrait(string identity, string key, object value)
         {
             try
             {
-                if (!(value is bool) && !(value is int) && !(value is string))
+                if (!(value is bool) && !(value is int) && !(value is string) && value != null)
                 {
-                    throw new ArgumentException("Value parameter must be string, int or boolean");
+                    throw new ArgumentException("Value parameter must be string, int, boolean or null");
```
gagantrivedi, line 242: I think traits supports `float`  as well?

matthewelwell: Yep, good shout. Have added.

gagantrivedi (dismissed):

matthewelwell:
Note that this is waiting on the release of: https://github.com/Flagsmith/flagsmith/pull/916.

FlagsmithClient.cs
```diff
@@ -261,6 +262,19 @@ public async Task<Trait> SetTrait(string identity, string key, object value)
...
+        public async Task SetTraits(string identity, Dictionary<string, object> traits)
+        {
+            ArrayList traitDataObjects = new ArrayList();
+
+            foreach (KeyValuePair<string, object> trait in traits)
+            {
+                traitDataObjects.Add(new { identity = new { identifier = identity }, trait_key = trait.Key, trait_value = trait.Value });
+            }
+
+            string url = configuration.ApiUrl.AppendPath("traits/bulk");
```
gagantrivedi, line 274: Maybe we should add a note that this endpoint is not part of edge and will not exist in v2

gagantrivedi (approved):

PR #28, opened 2022-05-25
Fix segment priorities

matthewelwell:


Flagsmith.Engine/Feature/Models/FeatureModel.cs
```diff
@@ -13,5 +13,20 @@ public class FeatureModel
         public string Type { get; set; }
         [JsonProperty(PropertyName = "name")]
         public string Name { get; set; }
+
+        public override bool Equals(object obj)
```
gagantrivedi, line 17: Why did we add this?

matthewelwell: Good question. I added that because I thought it was used when checking the equality of keys in a dictionary. Turns out (obviously) it uses the `getHashCode()` method. But since, I'd added it already, I figured I might as well leave it. I'm happy to remove it though if we think it's better?

gagantrivedi: Eh, either way is fine.

gagantrivedi (approved):

PR #29, opened 2022-05-27
Add semver support

matthewelwell:


Flagsmith.EngineTest/EngineTest.cs
```diff
@@ -31,7 +31,18 @@ public void Test_Engine(EnvironmentModel environmentModel, IdentityModel Identit
             Assert.Equal(sortedApiFlags.Count(), sortedEngineflags.Count());
             for (int i = 0; i < sortedEngineflags.Count(); i++)
             {
-                Assert.Equal(sortedApiFlags[i].feature_state_value?.ToString(), sortedEngineflags[i].GetValue(IdentityModel.DjangoId?.ToString())?.ToString());
+                var valueFromApi = sortedApiFlags[i].feature_state_value?.ToString();
+                var valueFromEngine = sortedEngineflags[i].GetValue(IdentityModel.DjangoId?.ToString())?.ToString();
+
+                if (valueFromApi == null)
+                {
+                    Assert.Null(valueFromEngine);
```
matthewelwell, line 39: I've separated this out as I was getting random failures. It seems that in c# `Assert.Equal(null, null)` does not always return true.

PR #39, opened 2022-06-16
Fix Flagsmith client trait handling

matthewelwell:


Flagsmith.FlagsmithClient/Trait.cs
```diff
@@ -6,63 +6,31 @@ namespace Flagsmith
...
         }
 
-        public int GetIntValue()
+        public string getTraitKey()
         {
-            return intValue;
+            return this.traitKey;
         }
 
-        public override string ToString()
```
gagantrivedi, line 63: Is this idiomatic c# to have public method use camel case instead of Pascal case?(not even sure if it's a method, though)

gagantrivedi (approved):
One minor comment/query

Flagsmith.FlagsmithClient/Trait.cs
matthewelwell: Good question. I think you're right. I will use Pascal case.

gagantrivedi (approved):

PR #44, opened 2022-07-12
Use feature name instead of id

matthewelwell:


.github/workflows/formatting-and-tests.yml
```diff
@@ -57,5 +57,9 @@ jobs:
                 dotnet-version: 6.0.x      
 
           - name: Run Tests
-            run: dotnet test
+            run: |
+              echo "Dotnet version: "
+              dotnet --version
```
matthewelwell, line 62: I have no idea how this managed to get the tests passing but I cannot get them to fail locally at all so I'm going to assume it's some GH weirdness and ignore!

matthewelwell: Sure! Can't hurt, right? RIGHT? 😬

gagantrivedi (approved):

PR #47, opened 2022-09-22
Interfaces for .net client

radek6733:


radek6733:
this is first pull request, i'm planning to do one more with the rest of discussed changes.

Flagsmith.FlagsmithClient/Flags.cs
```diff
@@ -34,17 +36,19 @@ public async Task<Flag> GetFlag(string featureName)
                 await _AnalyticsProcessor.TrackFeature(flag.GetFeatureName());
             return flag;
         }
-        public List<Flag> AllFlags() => _Flags;
+
+        public Task<IReadOnlyCollection<IFlag>> AllFlags() => Task.FromResult<IReadOnlyCollection<IFlag>>(_Flags);
+
         private static Flag FromFeatureStateModel(FeatureStateModel featureStateModel, string identityId = null) =>
```
matthewelwell, line 42: Since we're major versioning, I'd like to take the opportunity to clear up a few things here that I missed previously actually. For example, this (singular) FromFeatureStateModel should exist on the `Flag` class, not on the `Flags` class. 

These classes should mimic the models here: https://github.com/Flagsmith/flagsmith-python-client/blob/main/flagsmith/models.py. 

... unless there's a valid reason not to do this in C#.

Flagsmith.FlagsmithClient/Flags.cs
```diff
@@ -34,17 +36,19 @@ public async Task<Flag> GetFlag(string featureName)
...
-            new Flag(new Feature(featureStateModel.Feature.Name, featureStateModel.Feature.Id), featureStateModel.Enabled, featureStateModel.GetValue(identityId)?.ToString(), featureStateModel.Feature.Id);
+            new Flag(new Feature(featureStateModel.Feature.Name, featureStateModel.Feature.Id), featureStateModel.Enabled, featureStateModel.GetValue(identityId)?.ToString());
 
-        public static Flags FromFeatureStateModel(AnalyticsProcessor analyticsProcessor, Func<string, Flag> defaultFlagHandler, List<FeatureStateModel> featureStateModels, string identityId = null)
+        public static Flags FromFeatureStateModel(AnalyticsProcessor analyticsProcessor, Func<string, IFlag> defaultFlagHandler, List<FeatureStateModel> featureStateModels, string identityId = null)
```
matthewelwell, line 45: On the same logic as the above suggestion, this should be `FromFeatureStateModels`.

Flagsmith.FlagsmithClient/Flags.cs
```diff
@@ -34,17 +36,19 @@ public async Task<Flag> GetFlag(string featureName)
...
         {
             var flags = featureStateModels.Select(f => FromFeatureStateModel(f, identityId)).ToList();
             return new Flags(flags, analyticsProcessor, defaultFlagHandler);
         }
-        public static Flags FromApiFlag(AnalyticsProcessor analyticsProcessor, Func<string, Flag> defaultFlagHandler, List<Flag> flags)
-        => new Flags(flags, analyticsProcessor, defaultFlagHandler);
 
+        public static Flags FromApiFlag(AnalyticsProcessor analyticsProcessor, Func<string, IFlag> defaultFlagHandler, List<Flag> flags) =>
```
matthewelwell, line 51: This one should be `FromApiFlags`

Flagsmith.FlagsmithClient/FlagsmithClient.cs
```diff
@@ -28,217 +28,190 @@ namespace Flagsmith
...
-            ILogger logger = null,
-            Func<string, Flag> defaultFlagHandler = null,
-            bool enableAnalytics = false,
-            bool enableClientSideEvaluation = false,
-            int environmentRefreshIntervalSeconds = 60,
-            Dictionary<string, string> customHeaders = null,
-            int retries = 1,
-            double? requestTimeout = null,
-            HttpClient httpClient = null)
+        public FlagsmithClient(ILogger<FlagsmithClient> logger, IFlagsmithClientConfig config, HttpClient client = null)
```
matthewelwell, line 44: It seems _slightly_ odd to me that the logger is the first argument to this constructor but perhaps I'm not familiar enough with C#.

Flagsmith.FlagsmithClient/FlagsmithClient.cs
```diff
@@ -28,217 +28,190 @@ namespace Flagsmith
...
             {
                 throw new FlagsmithClientError("Local evaluation required to obtain identity segments.");
             }
-            IdentityModel identityModel = new IdentityModel { Identifier = identifier, IdentityTraits = traits?.Select(t => new TraitModel { TraitKey = t.GetTraitKey(), TraitValue = t.GetTraitValue() }).ToList() };
-            List<SegmentModel> segmentModels = Evaluator.GetIdentitySegments(
-                this.Environment, identityModel, new List<TraitModel>()
-            );
-
-            return segmentModels?.Select(t => new Segment(id: t.Id, name: t.Name)).ToList();
+            var identityModel = new IdentityModel { Identifier = identifier, IdentityTraits = traits?.Select(t => new TraitModel { TraitKey = t.Key, TraitValue = t.Value }).ToList() };
```
matthewelwell, line 103: What's the benefit to being less explicit with the variable typing here?

Flagsmith.FlagsmithClient/FlagsmithConfiguration.cs
```diff
@@ -1,59 +1,90 @@
...
+
         /// <summary>
         /// Additional headers to add to requests made to the Flagsmith API
         /// </summary>
         public Dictionary<string, string> CustomHeaders { get; set; }
-        public bool IsValid()
+
+        IReadOnlyDictionary<string, string> IFlagsmithClientConfig.CustomHeaders => CustomHeaders;
+
+        public static FlagsmithConfiguration From(
```
matthewelwell, line 65: What's the logic behind having this `From` method rather than a standard constructor?

Flagsmith.FlagsmithClient/Identity.cs
```diff
@@ -1,17 +1,22 @@
...
-    public class Identity
+    public class Identity : IIdentity
     {
         [JsonProperty]
         public List<Flag> flags;
 
         [JsonProperty]
         public List<Trait> traits;
 
+        IReadOnlyCollection<IFlag> IIdentity.Flags => flags;
```
matthewelwell, line 16: Apologies, this is likely my limited .NET knowledge but could you please explain what this is doing? Seems like we're perhaps overriding the `Flags` attributes on the `IIdentity` class but is there a reason we need to reference them as `IIdentity.Flags`?

Flagsmith.FlagsmithClient/Interfaces/IFlag.cs
```diff
@@ -0,0 +1,18 @@
...
+namespace Flagsmith.Interfaces
+{
+    public interface IFlag
+    {
+        int Id { get; }
+        bool Enabled { get; }
+        string Value { get; }
+        IFeature Feature { get; }
+
+        [Obsolete("Use 'Feature.Id' property, method will be removed in next version.")]
```
matthewelwell, line 12: Since we're creating a new major version here anyway, we can remove these methods and 'Obsolete' statements, right?

Flagsmith.FlagsmithClient/Interfaces/ITrait.cs
```diff
@@ -0,0 +1,16 @@
+﻿using System;
+
+namespace Flagsmith.Interfaces
+{
+    public interface ITrait
+    {
+        string Key { get; }
+        dynamic Value { get; }
+
+        [Obsolete("Use 'Key' property, method will be removed in next version.")]
```
matthewelwell, line 10: Again, we could probably remove these methods / obsolete statements?

matthewelwell (commented):
Thanks very much for this @radek6733 , it's really great! I've added a few comments although most of them are questions about why certain things have been done particular ways, likely most of them can be explained by my limited knowledge of .NET :)

Flagsmith.FlagsmithClient/Flags.cs
radek6733: sure, np. but i would prefer to make such changes after following PR (not created yet) where are many many more things :)

Flagsmith.FlagsmithClient/Flags.cs
radek6733: same here

Flagsmith.FlagsmithClient/Flags.cs
matthewelwell: Yes, that makes sense 👍 you can leave it to me in fact if you would like to.

Flagsmith.FlagsmithClient/FlagsmithClient.cs
radek6733: no worry here, it will be constructed by DI, all needed dependencies (constructor parameters) will be automatically constructed/reused. you can check web project (it might be in following PR)

Flagsmith.FlagsmithClient/FlagsmithClient.cs
radek6733: less typing :) the type is already in "new ....". using "var" is generally accepted and has also advantages when making refactors

Flagsmith.FlagsmithClient/FlagsmithConfiguration.cs
radek6733: yes can be constructor as well, can change it

Flagsmith.FlagsmithClient/Interfaces/ITrait.cs
radek6733: same here...

Flagsmith.FlagsmithClient/FlagsmithConfiguration.cs
matthewelwell: Unless there's a reason to do it this way then I think it makes more sense to use a constructor. That seems to be more universally accepted (since we're trying to maintain a consistent interface across our SDKs as much as possible).

Flagsmith.FlagsmithClient/Identity.cs
matthewelwell: Ah, ok, so am I right in thinking that you created the interface with the properties `Flags` and `Traits` to follow naming conventions. To avoid changing too many things, however, you've left the existing `flags` and `traits` fields on the `Identity` class and simply mapped the new `Flags` and `Traits` properties to the existing fields? 

Other than creating confusion, is there any reason why we can't define them like: 

```csharp
IReadOnlyCollection<IFlag> Flags => flags;
```

Sorry if I misunderstood your comment and you already answered this.

radek6733: it must be "public" (implicit implementation) or with "IIdentity." (explicit implementation). if public, it will be second definition for same reason (Flags vs flags), with explicit it wont be known to Identity class.

radek6733: The interface is also created for another reasons. To have immutable access to data in case you want to cache it (user cannot modify what is cached) and to allow different implementations or hide implementation details (like json serialization and so on)

matthewelwell: Ok, yep, that makes more sense now. Thanks for explaining.

matthewelwell:
@radek6733 I think the only outstanding change needed on this PR then is convert `FlagsmithConfiguration.From` to just a normal constructor. It also looks like the EngineTests might have broken so these will need fixing. These are quite complex though so if you need some help here then let me know and we can jump on a call.

matthewelwell (dismissed):

matthewelwell:
> Yp i think its because of the json data file, and I didnt make any changes to the engine. Its failing also in main branch now.

I don't see it failing in the main branch? I've just tested locally and it works fine for me. I've also just re-run the jobs on the release/5.0.0 PR and they've all passed. 

I can have a look at your PR to see if I can understand why the tests are failing.

matthewelwell:
Hmm... it seems that perhaps the tests in the Github workflows are a bit flaky. They're passing locally and I've got them to pass in Github by re-running them. I'll have to investigate this further but for now, I'll approve this and we can merge it.

matthewelwell (approved):

PR #55, opened 2022-11-02
Other changes

radek6733:


TGrannen:
I was doing some exploratory testing of Flagsmith and found that the Analytics upload was being awaited at the time of getting any given flag. This was very concerning to me as the performance of my application could be impacted if Flagsmith was having an issue. However, after looking through this PR, I noticed that this pattern would be removed. 

Is there a timeline for these changes to be released or should I just assume that it's not a good idea to turn on analytics for now?

matthewelwell:
Closed as stale.

PR #61, opened 2023-03-29
Feature/use interfaces

tberger:
Changed the use of the concrete implementation with the abstractions. Unit testing will be much easier when using the Flagsmith client.


matthewelwell (approved):
Thanks @tberger !

PR #68, opened 2023-06-13
chore(engine): add missing model properties

luk355:
Add missing properties to engine model so that engine flagState result can be mapped to API /identities response.


luk355:
Fair enough. Thanks for the comment. I have found a way how to lookup data from the DB model so this PR is not needed.

PR #72, opened 2023-07-07
feat: implement `IN` operator

khvn26:


Flagsmith.Engine/Segment/Evaluator.cs
```diff
@@ -53,16 +53,16 @@ static bool TraitsMatchSegmentCondition(List<TraitModel> identityTraits, Segment
 
         public static bool MatchesTraitValue(object traitValue, SegmentConditionModel condition)
         {
-            var exceptionOperatorMethods = new Dictionary<string, string>()
-            {
-                { Constants.NotContains, "EvaluateNotContains" },
-                { Constants.Regex, "EvaluateRegex" },
-                { Constants.Modulo, "EvaluateModulo" },
```
matthewelwell, line 60: I'd rather revert the formatting change here if possible. Not 100% sure if this is correct formatting for c# but it's currently consistent throughout the codebase so we should try to keep it that way (or submit a PR to change it everywhere if we think it's wrong).

matthewelwell (changes requested):

Flagsmith.Engine/Segment/Evaluator.cs
```diff
@@ -53,16 +53,16 @@ static bool TraitsMatchSegmentCondition(List<TraitModel> identityTraits, Segment
...
+                {Constants.In, "EvaluateIn"},
+                {Constants.NotContains, "EvaluateNotContains"},
+                {Constants.Regex, "EvaluateRegex"},
+                {Constants.Modulo, "EvaluateModulo"},
             };
 
             if (exceptionOperatorMethods.TryGetValue(condition.Operator, out var operatorMethod))
             {
-                return (bool)typeof(SegmentConditionModel).GetMethod(operatorMethod).Invoke(condition, new object[] { traitValue });
+                return (bool)typeof(SegmentConditionModel).GetMethod(operatorMethod).Invoke(condition, new object[] { traitValue.ToString() });
```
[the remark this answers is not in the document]
khvn26: > Based on this, we should probably change the signature of the EvaluateModulo method?

Done.

> Perhaps we should also make it obvious that we're casting the traitValue to double (pull it to the top of the method), but maybe that's less important?

I think it's pretty obvious as it is.

Flagsmith.Engine/Segment/Evaluator.cs
khvn26: Sure, done.
Any tips for setting up auto-formatting to support the common style?

matthewelwell: Sorry, forgot to respond here. I don't have anything built in actually. I just run `dotnet format` manually I think... we should add this to a pre-commit hook.

matthewelwell (approved):

PR #77, opened 2023-10-05
Add caching

JFCote:
This is a tentative to handle caching.

Since Flagsmith now works with a list of flag instead of "flag by flag", normal cache are not working for this scenario.
I have created simple cache classes that handle if the data is stale or not.

I did my best with the identity and traits but its gets pretty complicated with caching.

Please let me know what is missing and I'll try to provide everything for a quick merge because we need this feature :)


JFCote:
@dabeeeenster Please let me know if something is missing.

dabeeeenster:
cc @matthewelwell for review

JFCote:
Can you guys consider adding the "hacktoberfest" to the project label? This will allow this pr to count towards my goal of 4 pr during the month of October if it's merged before October 31th.

But if you don't,  that's OK too,  it's just a bonus of you do! Thanks!!

dabeeeenster:
Done!

JFCote:
@dabeeeenster @matthewelwell any news when you'll be able to review this? I have time this week to collaborate and add stuff if required but after that I will not.
Thanks!

matthewelwell:
Hey @JFCote, apologies, I've just got back from some time off - I will take a look this week.

Flagsmith.FlagsmithClient/Cache/IdentityFlagListCache.cs
```diff
@@ -0,0 +1,26 @@
...
+    {
+        private readonly IdentityTraitsKey _identityTraitsKey;
+
+        public IdentityFlagListCache(IdentityTraitsKey identityTraitsKey, IFlags flags, IDateTimeProvider dateTimeProvider, int cacheDurationInMinutes) :
+            base(dateTimeProvider, flags, cacheDurationInMinutes)
+        {
+            _identityTraitsKey = identityTraitsKey;
+        }
+
+        public IFlags GetLatestFlags(GetIdentityFlagsDelegate getFlagsDelegate)
```
matthewelwell, line 15: Perhaps this is my limited knowledge of C# but I'm a bit confused here as to why we're passing in the GetIdentityFlagsDelegate object that seems to already be defined on the abstract class? This seems like it could cause unexpected side effects. Could you explain the logic here and what this Delegate object's intended behaviour is?

Flagsmith.FlagsmithClient/Cache/IdentityTraitsKey.cs
```diff
@@ -0,0 +1,38 @@
...
+using System.Text;
+
+namespace Flagsmith
+{
+    public class IdentityTraitsKey
+    {
+        public string Identity { get; set; }
+        public List<ITrait> Traits { get; set; }
+
+        public IdentityTraitsKey(string identity, List<ITrait> traits)
```
matthewelwell, line 12: We should use `identifier` whenever we're talking about the string key that represents an identity.

Flagsmith.FlagsmithClient/Cache/IdentityTraitsKey.cs
```diff
@@ -0,0 +1,38 @@
...
+
+                var stringBuilder = new StringBuilder();
+                foreach (var t in hashBytes)
+                {
+                    stringBuilder.Append(t.ToString("x2"));
+                }
+
+                return stringBuilder.ToString();
+            }
+        }
```
matthewelwell, line 36: As a python developer, this seems overly verbose but perhaps there is no better way than this in C#?

Flagsmith.FlagsmithClient/FlagsmithClient.cs
```diff
@@ -74,7 +81,9 @@ public class FlagsmithClient : IFlagsmithClient
             Dictionary<string, string> customHeaders = null,
             int retries = 1,
             double? requestTimeout = null,
-            HttpClient httpClient = null)
+            HttpClient httpClient = null,
+            bool cachingEnabled = false,
+            int cacheDurationInMinutes = 5)
```
matthewelwell, line 86: I would rather properties like this were contained within a separate `CacheConfig` class for example to make it more easily extensible without having to change the constructor on the main client class.

Flagsmith.FlagsmithClient/FlagsmithClient.cs
```diff
@@ -113,7 +131,18 @@ public class FlagsmithClient : IFlagsmithClient
         /// Get all the default for flags for the current environment.
         /// </summary>
         public async Task<IFlags> GetEnvironmentFlags()
-            => Environment != null ? GetFeatureFlagsFromDocuments() : await GetFeatureFlagsFromApi();
+        {
+            if (CachingEnabled)
+            {
+                return _regularFlagListCache.GetLatestFlags(GetFeatureFlagsFromCorrectSource);
+            }
+            return await GetFeatureFlagsFromCorrectSource();
```
matthewelwell, line 139: Ok, I see the logic here. I tend to prefer a simpler interface on the caching side (if possible) and extend the code here slightly to something along the lines of: 

```csharp
IFlags _environmentFlags = null;
if (CachingEnabled) 
{
    _environmentFlags = cache.getEnvironmentFlags();
}
if (!_environmentFlags) {
    _environmentFlags = Environment != null ? GetFeatureFlagsFromDocuments() : await GetFeatureFlagsFromApi();  // maybe this could be a separate private method as you have
    if (CachingEnabled)
    {
        cache.setEnvironmentFlags(_environmentFlags);
    }
}
return _environmentFlags;
```

Maybe this is a bit verbose, with a few too many conditions so there may be some improvements that could be made but I would definitely prefer to keep the client logic out of the cache if possible. I appreciate your approach only passed in the method, but it still feels cleaner this way I think? Happy to discuss though.

Flagsmith.FlagsmithClient/Cache/IdentityTraitsKey.cs
```diff
@@ -0,0 +1,38 @@
...
+        public string Identity { get; set; }
+        public List<ITrait> Traits { get; set; }
+
+        public IdentityTraitsKey(string identity, List<ITrait> traits)
+        {
+            Identity = identity;
+            Traits = traits;
+        }
+
+        public string GenerateUniqueKey()
```
matthewelwell, line 18: I wonder if this could just be included on a wrapper class for the Identity itself somehow?

matthewelwell (commented):
thanks for this @JFCote. I've added a few comments here but I'd be happy to discuss further as I appreciate that I may not understand some of the language specific patterns.

Flagsmith.FlagsmithClient/Cache/IdentityTraitsKey.cs
JFCote: will fix that shortly

Flagsmith.FlagsmithClient/Cache/IdentityTraitsKey.cs
JFCote: This is kind of the idea for the `IdentityTraitsKey` class that we are in. It is used to identify a unique combinaison of the identity + the specific traits. Or am I not understanding your question?

Flagsmith.FlagsmithClient/FlagsmithClient.cs
JFCote: Well this is the idea of the delegate. Let me explain and maybe we can find the best approach.

I don't think your approach would work because it means your client is currently calling the `GetFeatureFlagsFromApi()` every time we want the environment flags. Even with your first condition, this might return stale / outdated flags.

My idea is that the cache job is to know if the values are stale/outdated. It's almost it's only job. But it doesn't know how to get the new value. Enters the concept of delegate in C#. It's something similar as "passing a function" if you did good old C++ but much more secure. The idea is that you define a signature that is expected for a method to be called using the keyword "delegate". That way, the cache only get passed `GetIdentityFlagsDelegate`. The cache don't know what this method do, only that when it detects that the cache is stale/outdated, it calls this method (which comes from the client) and replace the old values with the new one.

This has the advantages of:
* Keeping the logic of "client" that has always been done in the client, in the FlagsmithClient.
* Keeping the job of the cache in the cache.

On a sidenote, cache are usually much more easier to do than that and there are a lot of library built for this in C# and .NET but they work only as key/value dictionary. In your API design, environment flags are now returned in batch (all of them) which do not work with how default caching libraries works (they would work perfectly with a design where you go search for each single flag, when required, like it was in Flagsmith a couple of years back).

That being said, that's why I created these "manual" cache but I'm pretty sure they will be as powerful and they are very simple :)
Let me know if something is not clear!

Flagsmith.FlagsmithClient/Cache/IdentityFlagListCache.cs
JFCote: You can see my comment here where I explain what is the goal of a delegate: https://github.com/Flagsmith/flagsmith-dotnet-client/pull/77/files#r1363795518

Let me know if you need more information

JFCote:
@matthewelwell Updated the problem your found. Let me know if there is anything else!

Flagsmith.FlagsmithClient/FlagsmithClient.cs
matthewelwell: >I don't think your approach would work because it means your client is currently calling the GetFeatureFlagsFromApi() every time we want the environment flags. Even with your first condition, this might return stale / outdated flags.

I don't _think_ this is true, is it? If the cache contains something, it won't enter the second conditional block. 

We can even simplify the calls to the cache in my suggestion above to just: 

```csharp
_environmentFlags = cache.get(environmentKey);
```

(assuming there aren't any collisions between the environmentKey and however we hash identities).

>In your API design, environment flags are now returned in batch (all of them) which do not work with how default caching libraries works (they would work perfectly with a design where you go search for each single flag, when required, like it was in Flagsmith a couple of years back)

Another solution to this might be to simply write a `hash` method for the `Flags` class, right?

Flagsmith.FlagsmithClient/Cache/IdentityTraitsKey.cs
JFCote: Who will be using this wrapper? Not sure to understand, sorry :)

Flagsmith.FlagsmithClient/FlagsmithClient.cs
matthewelwell: I mean, I was expecting that the `cache.get` would of course have logic in it to determine if it's stale and only return something if it's not stale.

Flagsmith.FlagsmithClient/Cache/IdentityTraitsKey.cs
matthewelwell: Only our internal code right now would use this, but it seems more extensible to me than creating a separate class solely for the purpose of holding a unique key?

Flagsmith.FlagsmithClient/FlagsmithClient.cs
JFCote: This is where I think we don't understand each other. But don't worry, I will try to explain myself better!

The work of this kind of cache is to **always** return something. You either use the cache or don't use it. When you use it, the cache always returns something and you don't have to worry if the data is stale or not. This is the job of the cache. When you request the content of the cache, if the cache detect that the data is old, it will take care of refreshing the data before sending back to you. That way, the client just have to know to use or not to use the cache and to provide a way for the cache to update itself if required (the delegate).

The type of cache you propose in your code above is an "auto-clean" cache (for a lack of better words). It means the cache will clean itself when it becomes stale. We could use a cache like this but it would require much more work and it puts the job to update it in the hands of the client, which is weird.
Why more work? The cache needs to have scheduled task avec X time to check if the stuff inside if stale. If it is, delete it. Then, when someone calls the cache, it returns empty and the client must fill the cache itself which should definitely not be its job.

In my code, the cache is 100% independent and the only thing it requires is a "pointer" to a method to know how to update itself.

Flagsmith.FlagsmithClient/Cache/IdentityTraitsKey.cs
JFCote: all right, got you! Will try to come up with something better today! :)

Flagsmith.FlagsmithClient/FlagsmithClient.cs
matthewelwell: I can certainly see what you're saying, and I am not against the approach you have taken. But just to clarify regarding your comment here: 

>The type of cache you propose in your code above is an "auto-clean" cache (for a lack of better words). It means the cache will clean itself when it becomes stale. We could use a cache like this but it would require much more work and it puts the job to update it in the hands of the client, which is weird.
> Why more work? The cache needs to have scheduled task every X time to check if the stuff inside if stale. If it is, delete it. Then, when someone calls the cache, it returns empty and the client must fill the cache itself which should definitely not be its job.

What I mean is just that the cache stores 2 things for each key: the cached data, and the datetime it was last updated. Then, when you call `cache.get(key)` it checks if the cache is still valid (based on some configured ttl) and only returns the data if it is, if it's not, it will clear the key and return an empty result. As far as I can tell, this achieves almost exactly what your code achieves without the need for passing the method pointer or having to care about whether the user is requesting environment flags or identity flags. This way, the cache is still 'dumb' and only cares about storing data and how long it is valid for.

JFCote: You are correct. But if we go that way, this means the client will have to fill the cache, which should not be its job in my opinion. But if it is what you want, I can modify the code towards that goal.

Flagsmith.FlagsmithClient/FlagsmithClient.cs
```diff
@@ -95,6 +103,13 @@ public class FlagsmithClient : IFlagsmithClient
...
+            if (CacheConfig.Enabled)
+            {
+                _regularFlagListCache = new RegularFlagListCache(new DateTimeProvider(),
+                    GetEnvironmentFlags().Result,
+                    CacheConfig.DurationInMinutes);
+                _flagListCacheDictionary = new Dictionary<string, IdentityFlagListCache>();
+            }
         }
 
         /// <summary>
```
vpetrusevici, line 115: I think we should also add  CacheConfig in IFlagsmithConfiguration and in this constructor.

vpetrusevici: rows 120-123

JFCote: Added!!

Flagsmith.FlagsmithClient/FlagsmithClient.cs
JFCote: Let me know if anything needs to be changed here, thanks!

JFCote:
@matthewelwell @vpetrusevici Please let me know if any changes are required. Our company is a paying customer for Flagsmith, working mainly in C# and .NET and this caching is very important for us to continue our project. I'm available to update / add anything that is required to merge this quickly.
Thanks!

Flagsmith.FlagsmithClient/FlagsmithClient.cs
matthewelwell: Hi @JFCote apologies for the delay. As a pattern, the other cache implementations do follow a structure more like what I have suggested, that being said, since it's invisible to the user it doesn't matter too much. In the not too distant future, I would like to build a consistent caching solution across all our server side clients but that's another conversation! 

I would definitely prefer the approach that I have suggested here but I'm happy to go with the one included in this PR for now, since changing it in the future shouldn't be too much trouble.

matthewelwell (approved):

matthewelwell:
@JFCote I've responded to / resolved all your comments and approved the PR. Let me know if there is any reason not to merge and release.

JFCote:
@matthewelwell I see no reason to prevent the merge and the release. I will be able to test it in real scenario and come back with tweaks, bug fixes, etc..
I tend to be a little bit more active in this project in the future, if that is ok with you guys at Flagsmith :)
Have a wonderful day and let me know once it's released!

PR #81, opened 2023-10-26
Fix cache initialisation that was some kind of circular reference.

JFCote:
There was no need to initiate the cache right at the beginning. 
In fact, it was creating a problem trying to call `GetLatestFlags` on the cache that was currently trying to be initialized for the first time.

### DIFF
--- a/Flagsmith.FlagsmithClient/Cache/FlagListCache.cs
+++ b/Flagsmith.FlagsmithClient/Cache/FlagListCache.cs
@@ -14,9 +14,9 @@ namespace Flagsmith.Cache
         protected readonly IDateTimeProvider _dateTimeProvider;
         protected DateTime? _timestamp;
 
-        protected FlagListCache(IDateTimeProvider dateTimeProvider, IFlags flags, int cacheDurationInMinutes)
+        protected FlagListCache(IDateTimeProvider dateTimeProvider, int cacheDurationInMinutes)
         {
-            _flags = flags;
+            _flags = null;
             _dateTimeProvider = dateTimeProvider;
             _timestamp = null;
             _cacheDurationInMinutes = cacheDurationInMinutes;
@@ -24,12 +24,9 @@ namespace Flagsmith.Cache
 
         protected bool IsCacheStale()
         {
-            if (_timestamp == null)
-            {
-                return true;
-            }
-
-            return _dateTimeProvider.Now().Subtract((DateTime)_timestamp).TotalMinutes > _cacheDurationInMinutes;
+            return _timestamp == null ||
+                   _flags == null ||
+                   _dateTimeProvider.Now().Subtract((DateTime)_timestamp).TotalMinutes > _cacheDurationInMinutes;
         }
     }
 }
--- a/Flagsmith.FlagsmithClient/Cache/FlagListCache.cs
+++ b/Flagsmith.FlagsmithClient/Cache/FlagListCache.cs
@@ -18,7 +18,10 @@ namespace Flagsmith.Cache
         {
             _flags = flags;
             _dateTimeProvider = dateTimeProvider;
-            _timestamp = dateTimeProvider.Now();
+
+            //This is to ensure that the cache is stale on first run
+            _timestamp = dateTimeProvider.Now().Subtract(new TimeSpan(0, 0, cacheDurationInMinutes + 1, 0));
+
             _cacheDurationInMinutes = cacheDurationInMinutes;
         }
 
--- a/Flagsmith.FlagsmithClient/Cache/FlagListCache.cs
+++ b/Flagsmith.FlagsmithClient/Cache/FlagListCache.cs
@@ -12,22 +12,24 @@ namespace Flagsmith.Cache
 
         protected IFlags _flags;
         protected readonly IDateTimeProvider _dateTimeProvider;
-        protected DateTime _timestamp;
+        protected DateTime? _timestamp;
 
         protected FlagListCache(IDateTimeProvider dateTimeProvider, IFlags flags, int cacheDurationInMinutes)
         {
             _flags = flags;
             _dateTimeProvider = dateTimeProvider;
-
-            //This is to ensure that the cache is stale on first run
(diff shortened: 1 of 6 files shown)


JFCote:
@matthewelwell Here is a tentative fix for a problem on initialization I have using the cache in my code.
Let me know if you have any questions.

Flagsmith.FlagsmithClient/Cache/FlagListCache.cs
```diff
@@ -18,7 +18,10 @@ protected FlagListCache(IDateTimeProvider dateTimeProvider, IFlags flags, int ca
         {
             _flags = flags;
             _dateTimeProvider = dateTimeProvider;
-            _timestamp = dateTimeProvider.Now();
+
+            //This is to ensure that the cache is stale on first run
+            _timestamp = dateTimeProvider.Now().Subtract(new TimeSpan(0, 0, cacheDurationInMinutes + 1, 0));
```
matthewelwell, line 23: This feels a bit hacky, can't we just allow `_timestamp` to be null and initialise it as such?

Flagsmith.FlagsmithClient/FlagsmithClient.cs
```diff
@@ -96,17 +96,20 @@ public class FlagsmithClient : IFlagsmithClient
...
             {
                 _pollingManager = new PollingManager(GetAndUpdateEnvironmentFromApi, EnvironmentRefreshIntervalSeconds);
                 _ = _pollingManager.StartPoll();
             }
+
             if (CacheConfig.Enabled)
             {
                 _regularFlagListCache = new RegularFlagListCache(new DateTimeProvider(),
-                    GetEnvironmentFlags().Result,
+                    new Flags(new List<IFlag>(), _analyticsProcessor, DefaultFlagHandler),
```
matthewelwell, line 112: I don't really understand the logic here - can't we just add a constructor on the `RegularFlagListCache` class that doesn't include the Flags and initialise it there or as null perhaps? This feels a little bit hacky again, having to initialise an empty flags list for no real reason.

Flagsmith.FlagsmithClient/Cache/FlagListCache.cs
JFCote: I was trying to be quick. You are right, I can find something better than that. :)
Will fix and submit something today!

Flagsmith.FlagsmithClient/FlagsmithClient.cs
JFCote: Yeah I totally understand... The problem is that the `Flags` object requires an `analyticsProcessor` and a `FlagHandler`. These must be injected but they are only known in the FlagsmithClient...
But I will find something that is more eleguant! Give me a couple of hours :)

JFCote: Fixed!

Flagsmith.FlagsmithClient/Cache/FlagListCache.cs
```diff
@@ -12,19 +12,24 @@ public abstract class FlagListCache
...
 
         protected bool IsCacheStale()
         {
-            return _dateTimeProvider.Now().Subtract(_timestamp).TotalMinutes > _cacheDurationInMinutes;
+            if (_timestamp == null)
+            {
+                return true;
+            }
+
+            return _dateTimeProvider.Now().Subtract((DateTime)_timestamp).TotalMinutes > _cacheDurationInMinutes;
```
matthewelwell, line 32: This is a personal (pythonic) preference, but I would rather see something like: 

```csharp
return _timestamp == null || _dateTimeProvider.Now().Subtract((DateTime)_timestamp).TotalMinutes > _cacheDurationInMinutes;
```

(maybe this is not the correct ternary form here but something like this seems cleaner in my opinion but this isn't necessary for approval)

JFCote:
@matthewelwell I updated the code and also added a test showing that initializing the cache is now working :)

Flagsmith.FlagsmithClient/Cache/IdentityFlagListCache.cs
```diff
@@ -1,13 +1,21 @@
...
 
-        public IdentityFlagListCache(IdentityWrapper identityWrapper, IFlags flags, IDateTimeProvider dateTimeProvider, int cacheDurationInMinutes) :
-            base(dateTimeProvider, flags, cacheDurationInMinutes)
+        public IdentityFlagListCache(IdentityWrapper identityWrapper,
+            IDateTimeProvider dateTimeProvider,
+            AnalyticsProcessor analyticsProcessor,
+            Func<string, IFlag> defaultFlagHandler,
+            int cacheDurationInMinutes) :
+            base(dateTimeProvider,
+                new Flags(new List<IFlag>(), analyticsProcessor, defaultFlagHandler),
```
matthewelwell, line 17: Can't we just make `flags` nullable in the base class and scrap all of this?

Flagsmith.FlagsmithClient/Cache/FlagListCache.cs
JFCote: Good idea! Will update this in a minute

Flagsmith.FlagsmithClient/Cache/IdentityFlagListCache.cs
JFCote: Yes I think it could work! Give me a minute!

Flagsmith.FlagsmithClient/Cache/FlagListCache.cs
JFCote: fixed

Flagsmith.FlagsmithClient/Cache/IdentityFlagListCache.cs
JFCote: fixed

JFCote:
@matthewelwell Let me know if anything else is required!

matthewelwell (approved):
Beautiful! Thanks for sticking with me there @JFCote.

matthewelwell:
@JFCote fix version is here: https://www.nuget.org/packages/Flagsmith/5.2.1

PR #83, opened 2023-10-29
fix: Ensure environment is retrieved on start polling

novakzaballa:
Ensures environment is retrieved on start polling by executing the StartPoll() method synchronoulsy, and executing the first call to the callback `GetAndUpdateEnvironmentFromApi` also asynchronoulsy and immediately. While(true) and Delay funcitons replaced by Timer class to allow the StartPoll() to be executed synchronously without hanging the execution of the main thread.

### DIFF
--- a/Flagsmith.FlagsmithClient/FlagsmithClient.cs
+++ b/Flagsmith.FlagsmithClient/FlagsmithClient.cs
@@ -242,7 +242,7 @@ namespace Flagsmith
                     }
 
                     var cancellationTokenSource = new CancellationTokenSource(TimeSpan.FromSeconds(RequestTimeout ?? 100));
-                    HttpResponseMessage response = await _httpClient.SendAsync(request, cancellationTokenSource.Token);
+                    HttpResponseMessage response = await _httpClient.SendAsync(request, cancellationTokenSource.Token).ConfigureAwait(false);
                     return response.EnsureSuccessStatusCode();
                 })).Content.ReadAsStringAsync();
             }
--- a/Flagsmith.FlagsmithClient/FlagsmithClient.cs
+++ b/Flagsmith.FlagsmithClient/FlagsmithClient.cs
@@ -103,7 +103,7 @@ namespace Flagsmith
             if (EnableClientSideEvaluation)
             {
                 _pollingManager = new PollingManager(GetAndUpdateEnvironmentFromApi, EnvironmentRefreshIntervalSeconds);
-                _ = _pollingManager.StartPoll();
+                _pollingManager.StartPoll().GetAwaiter().GetResult();
             }
 
             if (CacheConfig.Enabled)
--- a/Flagsmith.FlagsmithClient/PollingManager.cs
+++ b/Flagsmith.FlagsmithClient/PollingManager.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Runtime.InteropServices;
 using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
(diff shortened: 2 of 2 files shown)


Flagsmith.FlagsmithClient/FlagsmithClient.cs
```diff
@@ -103,7 +103,7 @@ public class FlagsmithClient : IFlagsmithClient
             if (EnableClientSideEvaluation)
             {
                 _pollingManager = new PollingManager(GetAndUpdateEnvironmentFromApi, EnvironmentRefreshIntervalSeconds);
-                _ = _pollingManager.StartPoll();
+                _pollingManager.StartPoll().GetAwaiter().GetResult();
```
matthewelwell, line 106: This looks a bit ugly - what's the reasoning here?

Flagsmith.FlagsmithClient/PollingManager.cs
```diff
@@ -21,19 +23,23 @@ public PollingManager(Func<Task> callBack, int intervalSeconds = 10)
...
-                await Task.Delay(_Interval, _CancellationTokenSource.Token);
                 if (_CancellationTokenSource.Token.IsCancellationRequested)
-                    break;
-            }
+                {
+                    _timer.Dispose();
+                    return;
+                }
+                await _CallBack.Invoke();
+            }, null, _Interval, _Interval);
```
matthewelwell, line 42: From what I can tell, [this](https://learn.microsoft.com/en-us/dotnet/api/system.threading.timer.-ctor?view=net-7.0#system-threading-timer-ctor(system-threading-timercallback-system-object-system-timespan-system-timespan)) is the relevant constructor for this Timer instance? It look me a bit of time to locate that though as it seems that the most common example of the `Timer` class is from `System.Timers` which only has 0/1 argument constructors. 

Perhaps Visual Studio / VSCode has some form of intellisense on this so a comment isn't necessary but I just wanted to highlight that it took me some time to find it - because I wasn't sure why we were providing the `_Interval` argument twice.

matthewelwell (commented):
Added a couple of minor comments but on the whole this looks like a very neat solution, nice work!

Flagsmith.FlagsmithClient/FlagsmithClient.cs
novakzaballa: May I ask why it looks a bit ugly? The reasoning is just to accomplish the issue objective:

> Ensures environment is retrieved on start polling by executing the StartPoll() method synchronously

Why use this particular way to make the call async? As per some tech docs I found, like this one https://devblogs.microsoft.com/dotnet/configureawait-faq/, this is the way the compiler would interpret an `await task ` code, and also, it would handle the error bubbling up an internal error happening in an internal node of a nested call, so we don't need to change our error handlers to use an error stacker as mentioned here https://gsferreira.com/archive/2020/08/avoid-getawaiter-getresult-at-all-cost/.

Making an async method work synchronously is strongly unrecommended in general, but this seems to be one of the few cases where it is needed since we want to call it in a constructor method, which by definition should execute synchronously since it should complete any initialization of the instance before any other action with the instance.

Flagsmith.FlagsmithClient/PollingManager.cs
novakzaballa: Well, it was easy for me to find this constructor, maybe, because I was looking for this particular constructor signature, since I wanted to execute the callback synchronously a first time, and then use the Timer to execute the callback every _Interval time, but delaying the first execution the same _Interval time to not repeat the first call.

matthewelwell (approved):

Issue #88, opened 2024-01-24
Exception: System.ThrowHelper.ThrowInvalidOperationException_ConcurrentOperationsNotSupported


ivastokic:
Hi,
We were getting this error in production environment on one of our pods:

![Screenshot 2024-01-24 at 14 26 26](https://github.com/Flagsmith/flagsmith-dotnet-client/assets/112628083/b55db674-4d14-4aea-b762-e933179efbfb)

It started yesterday evening, and restarting of the pod stopped it. We followed your example how to register service, also didn't have any related changes recently. Stack pointing to this one:

```
public async Task TrackFeature(string featureName)
    {
      int num;
      this.AnalyticsData[featureName] = this.AnalyticsData.TryGetValue(featureName, out num) ? num + 1 : 1;
      if ((DateTime.Now - this._LastFlushed).Seconds <= this._FlushIntervalSeconds)
        return;
      await this.Flush();
    } 
```
 Do you have any idea why this happened, could it be related to lack of thread safety?

We are using version 5.0.1

novakzaballa:
Hi @ivastokic. Could you please provide us with more information? For example:
Are you using .Net Core or .Net Framework? Which version?
Also, from the info and context provided, it looks like a safe thread issue, however, it is hard to say the root cause or recommend a solution without more context. Can you generate a minimal project reproducing the issue and your current configuration?

MaximTkachenko:
It looks like you have a thread sefety issue inside `AnalyticsProcessor`:
- according to ypur documentatyion `FlagsmithClient` is used as a sigletone
- it means that `AnalyticsProcessor` created inside `FlagsmithClient` is used as a sigleton too
- inside `AnalyticsProcessor` you use plain dictionary for feature tracking without any synchronization and periodically you call `Clear` for this dictionary

So you read and modify this dictionary concurrently without any protection. I suggest you to use [ReaderWriterLock ](https://learn.microsoft.com/en-us/dotnet/api/system.threading.readerwriterlock?view=net-8.0) or [CoincurrentDictionary](https://learn.microsoft.com/en-us/dotnet/api/system.collections.concurrent.concurrentdictionary-2?view=net-8.0).
![image](https://github.com/Flagsmith/flagsmith-dotnet-client/assets/1560515/ccdb3877-6b76-4aba-b1c9-07c990822d14)

novakzaballa:
[...]

PR #93, opened 2024-03-30
fix: Concurrency issue with analytics processor data

novakzaballa:
This PR fixes a potential concurrency issue with the dictionary `AnalyticsData` used in the AnalyticsProcessor class.
One test reproducing the issue has been added, and the mentioned dictionary has been turned into a `ConcurrentDictionary' which implements appopiate locking logic for concurrent access.

Please refer to this issue https://github.com/Flagsmith/flagsmith-dotnet-client/issues/88 for more context.

### DIFF
--- a/Flagsmith.FlagsmithClient/AnalyticsProcessor.cs
+++ b/Flagsmith.FlagsmithClient/AnalyticsProcessor.cs
@@ -8,6 +8,7 @@ using Newtonsoft.Json;
 using System.Threading;
 using Microsoft.Extensions.Logging;
 using Flagsmith.Extensions;
+using System.Collections.Concurrent;
 
 namespace Flagsmith
 {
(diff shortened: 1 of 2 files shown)


Flagsmith.FlagsmithClient/AnalyticsProcessor.cs
```diff
@@ -18,7 +19,7 @@ public class AnalyticsProcessor : IAnalyticsProcessor
         readonly string _EnvironmentKey;
         readonly int _TimeOut;
         DateTime _LastFlushed;
-        protected Dictionary<string, int> AnalyticsData;
+        protected ConcurrentDictionary <string, int> AnalyticsData;
```
MaximTkachenko, line 22: Though this switch to `ConcurrentDictionary` will fix the exception caused by concurrent operations on the plain dictionary, the logic here is still incorrect:
image
In the concurrent world you can't rely on simple increments like this `value = value + 1`, it does not give you a proper numbers. You need to rethink how you work with concurrency in this class.
Also you AnalyticsData is protected so other classes in your project might do some mess, it shiuld be private.

novakzaballa: While the original idea seems to be to avoid real parallel processing by using Flagsmith client as a Singleton and trying to orchestrate other components from there, currently that is not guaranteed for the TrackFeaturemethod of  AnalyticsProcessor class. I think that the best solution is refactoring the Analytics support multithreading, which I did in this PR https://github.com/Flagsmith/flagsmith-dotnet-client/pull/95.

PR #95, opened 2024-04-01
chore: Implement multi-threading support for analytics Processor

novakzaballa:
Implement support for multi-threading processing in the AnalyticsProcessor class.

Also, implements unit tests replicating potential issues. The issues show themselves only with a huge number of threads running concurrently, and at a certain number it is possible to reproduce them systematically, while that number could depend on the machine resources this has been tested on a MAC Book Pro M2 Processor with 16GB and on a PC with Corei7 and 48GB of RAM.

Please refer to this issue https://github.com/Flagsmith/flagsmith-dotnet-client/issues/88 for more context.

### DIFF
--- a/Flagsmith.Client.Test/AnalyticsTests.cs
+++ b/Flagsmith.Client.Test/AnalyticsTests.cs
@@ -102,59 +102,5 @@ namespace Flagsmith.FlagsmithClientTest
             // Then
             Assert.True(true);
         }
-
-        [Fact]
-        public async Task TestAnalyticsProcessorDataConcurrentConsistency()
-        {
-            // Given
-            var mockHttpClient = HttpMocker.MockHttpResponse(new HttpResponseMessage
-            {
-                StatusCode = System.Net.HttpStatusCode.OK,
-            });
-
-            Dictionary<string, int> featuresDictionary = new Dictionary<string, int>();
-            for (int i = 1; i <= 10; i++)
-            {
-                featuresDictionary.TryAdd($"Feature_{i}", 0);
-            }
-
-            var analyticsProcessor = new AnalyticsProcessorTest(mockHttpClient.Object, null, baseApiUrl: _defaultApiUrl);
-            const int numberOfThreads = 1000;
-            const int callsPerThread = 1000;
-
-            // When 
-            var tasks = new Task[numberOfThreads];
-            for (int i = 0; i < tasks.Length; i++)
-            {
-                var local = i;
-                string[] features = new string[callsPerThread];
-                for (int j = 0; j < callsPerThread; j++)
-                {
-                    string feature = $"Feature_{new Random().Next(1, featuresDictionary.Count + 1)}";
-                    features[j] = feature;
-                    featuresDictionary[feature]++;
-                }
-
-                tasks[i] = Task.Run(async () =>
-                {
-                    foreach (var feature in features)
-                    {
-                        await analyticsProcessor.TrackFeature(feature);
-                    }
-                });
-            }
-
-            await Task.WhenAll(tasks);
-
-            // Then
-            Dictionary<string, int> analyticsData = analyticsProcessor.GetAggregatedAnalytics();
-            int totalCallsMade = 0;
-            foreach (var feature in featuresDictionary)
-            {
-                totalCallsMade += analyticsData[feature.Key];
-                Assert.Equal(feature.Value, analyticsData[feature.Key]);
-            }
-            Assert.Equal(numberOfThreads * callsPerThread, totalCallsMade);
-        }
     }
 }
--- a/Flagsmith.Client.Test/AnalyticsTests.cs
+++ b/Flagsmith.Client.Test/AnalyticsTests.cs
@@ -17,10 +17,10 @@ namespace Flagsmith.FlagsmithClientTest
         public void TestAnalyticsProcessorTrackFeatureUpdatesAnalyticsData()
         {
             var analyticsProcessor = Fixtures.GetAnalyticalProcessorTest();
-            _ = analyticsProcessor.TrackFeature("myFeature");
-            Assert.Equal(1, analyticsProcessor["myFeature"]);
-            _ = analyticsProcessor.TrackFeature("myFeature");
-            Assert.Equal(2, analyticsProcessor["myFeature"]);
+            _ = analyticsProcessor.TrackFeature("TestAnalyticsProcessorTrackFeatureUpdatesAnalyticsDataFeature");
+            Assert.Equal(1, analyticsProcessor["TestAnalyticsProcessorTrackFeatureUpdatesAnalyticsDataFeature"]);
+            _ = analyticsProcessor.TrackFeature("TestAnalyticsProcessorTrackFeatureUpdatesAnalyticsDataFeature");
+            Assert.Equal(2, analyticsProcessor["TestAnalyticsProcessorTrackFeatureUpdatesAnalyticsDataFeature"]);
         }
         [Fact]
         public async Task TestAnalyticsProcessorFlushClearsAnalyticsData()
@@ -30,7 +30,7 @@ namespace Flagsmith.FlagsmithClientTest
                 StatusCode = System.Net.HttpStatusCode.OK,
             });
             var analyticsProcessor = new AnalyticsProcessorTest(mockHttpClient.Object, null, null);
-            await analyticsProcessor.TrackFeature("myFeature");
+            await analyticsProcessor.TrackFeature("TestAnalyticsProcessorFlushClearsAnalyticsDataFeature");
             await analyticsProcessor.Flush();
             Assert.False(analyticsProcessor.HasTrackingItemsInCache());
         }
@@ -42,13 +42,13 @@ namespace Flagsmith.FlagsmithClientTest
                 StatusCode = System.Net.HttpStatusCode.OK,
             });
             var analyticsProcessor = new AnalyticsProcessorTest(mockHttpClient.Object, null, baseApiUrl: _defaultApiUrl);
-            await analyticsProcessor.TrackFeature("myFeature1");
-            await analyticsProcessor.TrackFeature("myFeature2");
+            await analyticsProcessor.TrackFeature("TestAnalyticsProcessorFlushPostRequestDataMatchAnanlyticsDataFeature1");
+            await analyticsProcessor.TrackFeature("TestAnalyticsProcessorFlushPostRequestDataMatchAnanlyticsDataFeature2");
             var jObject = JObject.Parse(analyticsProcessor.ToString());
             await analyticsProcessor.Flush();
             mockHttpClient.verifyHttpRequest(HttpMethod.Post, "/api/v1/analytics/flags/", Times.Once);
-            Assert.Equal(1, jObject["myFeature1"].Value<int>());
-            Assert.Equal(1, jObject["myFeature2"].Value<int>());
+            Assert.Equal(1, jObject["TestAnalyticsProcessorFlushPostRequestDataMatchAnanlyticsDataFeature1"].Value<int>());
+            Assert.Equal(1, jObject["TestAnalyticsProcessorFlushPostRequestDataMatchAnanlyticsDataFeature2"].Value<int>());
         }
         [Fact]
         public async Task TestAnalyticsProcessorFlushEarlyExitIfAnalyticsDataIsEmpty()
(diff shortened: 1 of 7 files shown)


Flagsmith.Client.Test/FlagsmithTest.cs
```diff
@@ -431,5 +431,62 @@ public void TestCannotCreateFlagsmithClientInRemoteEvaluationWithoutAPIKey()
             var exception = Assert.Throws<Exception>(() => createFlagsmith());
             Assert.Equal("ValueError: environmentKey is required", exception.Message);
         }
+
+        [Fact]
+        public async Task TestAnalyticsDataConsistencyWithConcurrentCallsToGetFlags()
```
matthewelwell, line 436: This test seems to duplicate a lot of the test code in `TestAnalyticsProcessorDataConcurrentConsistency` - perhaps we can consolidate here?

Flagsmith.FlagsmithClient/FlagsmithClient.cs
```diff
@@ -408,6 +408,8 @@ private IFlags GetIdentityFlagsFromDocument(string identifier, List<ITrait>? tra
             return Flags.FromFeatureStateModel(_analyticsProcessor, DefaultFlagHandler, _engine.GetIdentityFeatureStates(Environment, identity), identity.CompositeKey);
         }
 
+        public Dictionary<string, int> aggregatedAnalytics => _analyticsProcessor != null ? _analyticsProcessor.GetAggregatedAnalytics() : new Dictionary<string, int>();
```
matthewelwell, line 411: Seems odd that this would return anything at all if `_analyticsProcessor` is null. Could this be typed as `Optional` and return null if `_analyticsProcessor` is null perhaps? Maybe there's a reason we want to return the empty dict though, happy to discuss.

Flagsmith.FlagsmithClient/AnalyticsProcessor.cs
```diff
@@ -79,9 +94,53 @@ public async Task Flush()
...
+
+            if (_LastFlushedInterval > _FlushIntervalSeconds)
                 await Flush();
         }
+
+        /// <summary>
+        /// Get aggregated analytics data.
+        /// </summary>
+        /// <returns>Dictionary of feature name and usage count</returns>
+        public Dictionary<string, int> GetAggregatedAnalytics()
```
matthewelwell, line 125: This needs some better documentation, or maybe it's sufficient (and better) to document all this behaviour on the class itself. It's quite complicated with the separate threads, etc.

Flagsmith.FlagsmithClient/AnalyticsProcessor.cs
```diff
@@ -79,9 +94,53 @@ public async Task Flush()
...
+        {
+            Dictionary<string, int> aggregatedAnalytics = new Dictionary<string, int>();
+            foreach (var threadCount in AnalyticsDataThreads.Values)
```
matthewelwell, line 128:
```suggestion
            foreach (var threadAnalyticsData in AnalyticsDataThreads.Values)
```

Flagsmith.FlagsmithClient/AnalyticsProcessor.cs
```diff
@@ -79,9 +94,53 @@ public async Task Flush()
...
+            {
+                foreach (var featureCount in threadCount)
```
matthewelwell, line 130:
```suggestion
                foreach (var trackedFeatureData in threadAnalyticsData)
```

matthewelwell (commented):
Overall this seems good, but I think some additional documentation would be really helpful in a number of places.

Flagsmith.Client.Test/AnalyticsTests.cs
```diff
@@ -16,10 +17,10 @@ public class AnalyticsTests
...
         {
             var analyticsProcessor = Fixtures.GetAnalyticalProcessorTest();
-            _ = analyticsProcessor.TrackFeature("myFeature");
-            Assert.Equal(1, analyticsProcessor["myFeature"]);
-            _ = analyticsProcessor.TrackFeature("myFeature");
-            Assert.Equal(2, analyticsProcessor["myFeature"]);
+            _ = analyticsProcessor.TrackFeature("TestAnalyticsProcessorTrackFeatureUpdatesAnalyticsDataFeature");
+            Assert.Equal(1, analyticsProcessor["TestAnalyticsProcessorTrackFeatureUpdatesAnalyticsDataFeature"]);
+            _ = analyticsProcessor.TrackFeature("TestAnalyticsProcessorTrackFeatureUpdatesAnalyticsDataFeature");
+            Assert.Equal(2, analyticsProcessor["TestAnalyticsProcessorTrackFeatureUpdatesAnalyticsDataFeature"]);
```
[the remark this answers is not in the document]
novakzaballa: I don't think the names are complicated, just long. The tests run in parallel, I used the name of the test as the name of the future to ensure the independence of each test.

Flagsmith.Client.Test/AnalyticsTests.cs
```diff
@@ -69,8 +70,91 @@ public async Task TestAnalyticsProcessorCallingTrackFeatureCallsFlushWhenTimerRu
...
+        [Fact]
+        public async Task TestAnalyticsProcessorDataConcurrentAccess()
+        {
+            // Given
+            var mockHttpClient = HttpMocker.MockHttpResponse(new HttpResponseMessage
+            {
+                StatusCode = System.Net.HttpStatusCode.OK,
+            });
+            var analyticsProcessor = new AnalyticsProcessorTest(mockHttpClient.Object, null, null);
+            const int numberOfCalls = 100000;
```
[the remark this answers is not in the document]
novakzaballa: The increase in the execution time of the tests on our CI platform is not a significant amount as you can see in the GH actions logs. If we reduce the amount I can not replicate systematically the issue in powerful machines like my PC or my Mac Book Pro with M2.

Flagsmith.Client.Test/AnalyticsTests.cs
```diff
@@ -69,8 +70,91 @@ public async Task TestAnalyticsProcessorCallingTrackFeatureCallsFlushWhenTimerRu
...
+                tasks[i] = Task.Run(async () =>
+                {
+                    await analyticsProcessor.TrackFeature($"Feature {local}");
+                });
+            }
+
+            await Task.WhenAll(tasks);
+
+            // Then
+            Assert.True(true);
```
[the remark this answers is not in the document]
novakzaballa: The only objective of this test is to verify that the code completes without any run time error, like the exception reported in this issue https://github.com/Flagsmith/flagsmith-dotnet-client/issues/88. If this test is applied to the current code base in the main branch, it fails as expected.

Flagsmith.Client.Test/AnalyticsTests.cs
```diff
@@ -69,8 +70,91 @@ public async Task TestAnalyticsProcessorCallingTrackFeatureCallsFlushWhenTimerRu
...
+
+            // When 
+            var tasks = new Task[numberOfThreads];
+            for (int i = 0; i < tasks.Length; i++)
+            {
+                var local = i;
+                string[] features = new string[callsPerThread];
+                for (int j = 0; j < callsPerThread; j++)
+                {
+                    string feature = $"Feature_{new Random().Next(1, featuresDictionary.Count + 1)}";
```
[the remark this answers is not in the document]
novakzaballa: Right, thank you!

Flagsmith.Client.Test/AnalyticsTests.cs
```diff
@@ -69,8 +70,91 @@ public async Task TestAnalyticsProcessorCallingTrackFeatureCallsFlushWhenTimerRu
...
+            }
+
+            await Task.WhenAll(tasks);
+
+            // Then
+            Assert.True(true);
+        }
+
+        [Fact]
+        public async Task TestAnalyticsProcessorDataConcurrentConsistency()
```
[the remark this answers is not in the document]
novakzaballa: I will remove this test and leave the one in higher level, at the client level. I will document that one.

Flagsmith.Client.Test/FlagsmithTest.cs
novakzaballa: I removed the test in the AnalyticsProcessor context as discussed and left only the one at the client level since they are duplicating the testing case and criteria.

Flagsmith.FlagsmithClient/AnalyticsProcessor.cs
novakzaballa: Done

Flagsmith.FlagsmithClient/AnalyticsProcessor.cs
novakzaballa: Done, thanks!

Flagsmith.FlagsmithClient/AnalyticsProcessor.cs
novakzaballa: Thanks, done

Flagsmith.Client.Test/Fixtures.cs
```diff
@@ -144,5 +143,36 @@ internal class Fixtures
...
+        {
+            get
+            {
+                List<object> flags = new List<object>();
+                // Add ten flags named Feature_1, Feature_2, ..., Feature_10
+                for (int i = 1; i <= 10; i++)
+                {
+                    object flag = new
+                    {
+                        id = 1,
```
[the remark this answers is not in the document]
novakzaballa: Done

Flagsmith.Client.Test/FlagsmithTest.cs
matthewelwell: I think the documentation could still be better here. Perhaps some comments throughout the test method itself to explain some of the nested loops for example.

matthewelwell (commented):
Added one extra comment, and I think one was missed from the previous review.

Flagsmith.FlagsmithClient/FlagsmithClient.cs
novakzaballa: The only reason behind this is because in general, I try to avoid returning results of different types in a function, since that would require additional logic in the calling code to manage those different types. In this case, since it is a Dictionary, iterating an empty Dictionary, or trying to find a specific key should be safe and will no require additional logic in the calling code.

Flagsmith.Client.Test/FlagsmithTest.cs
novakzaballa: Inline comments added. Please let me know your thoughts.

novakzaballa:
> Added one extra comment, and I think one was missed from the previous review.

Thanks! Both are answered now.

Flagsmith.Client.Test/FlagsmithTest.cs
```diff
@@ -431,5 +431,76 @@ public void TestCannotCreateFlagsmithClientInRemoteEvaluationWithoutAPIKey()
...
+                string[] features = new string[callsPerThread];
+
+                // Prepare an array of feature names of length callsPerThread.
+                for (int j = 0; j < callsPerThread; j++)
+                {
+                    // The feature names are randomly picked from the featuresDictionary which contains numberOfFeatures feature names.
+                    string featureName = $"Feature_{new Random().Next(1, featuresDictionary.Count + 1)}";
+                    features[j] = featureName;
+                    featuresDictionary[featureName]++;
+                }
```
matthewelwell, line 481:
```suggestion
                // Prepare an array of feature names of length callsPerThread.
                for (int j = 0; j < callsPerThread; j++)
                {
                    ```suggestion
                    // The feature names are randomly selected from the featuresDictionary and added to the 
                    // list of features, which represents the features that have been evaluated.  
                    string featureName = $"Feature_{new Random().Next(1, featuresDictionary.Count + 1)}";
                    features[j] = featureName;
                    
                    // The relevant key in the featuresDictionary is incremented to simulate an evaluation
                    // to track for that feature. 
                    featuresDictionary[featureName]++;
                }
```

Flagsmith.FlagsmithClient/FlagsmithClient.cs
matthewelwell: Ok, yep, makes sense.

Flagsmith.Client.Test/FlagsmithTest.cs
matthewelwell: I've added one suggested improvement.

matthewelwell (approved):

Issue #103, opened 2024-07-02
Providing an OfflineHandler causes the SDK to bypass live flag lookups (even if OfflineMode is false)


brendenk-cdata:
# Bug Report
**.NET SDK Version:** 5.3.0

## Description of problem
We're trying to make use of the offline handler feature to provide default flags if the SDK cannot contact our Flagsmith server for some reason. I.e., we're trying to use it for scenario 2 listed in this section of your documentation: https://docs.flagsmith.com/clients/server-side#using-an-offline-handler.

We have OfflineMode disabled, and have an instance of your `LocalFileHandler` class set as the offline handler. This is how we're instantiating the FlagsmithClient class when we register it with DI:
![image](https://github.com/Flagsmith/flagsmith-dotnet-client/assets/171082717/46fb3495-9d54-4f96-aa15-611345c72d6b)

## Expected Behavior
The SDK should still attempt live API calls to our Flagsmith server to obtain flag values, only using the offline handler as a fallback if the live API calls fail.

## Actual Behavior
The SDK never attempts any live API calls, it just uses the offline file directly.

## Analysis
The bug is easily reproducible because we can change the value of a flag on the Flagsmith server, but if we debug our code, the results of the `FlagsmithClient.GetEnvironmentFlags()` call shows us that the flag always has its offline file value.

Looking at the source code for `FlagsmithClient`, the cause becomes clear:
* `GetEnvironmentFlags()` calls `GetFeatureFlagsFromCorrectSource()`
* `GetFeatureFlagsFromCorrectSource()` is implemented to choose between a locally-loaded Environment model or a live API call based on whether the `Environment` variable is non-null: https://github.com/Flagsmith/flagsmith-dotnet-client/blob/4327e266a67f721c2c104ad5df4298d49b220436/Flagsmith.FlagsmithClient/FlagsmithClient.cs#L204-L207
* The `Environment` variable is always non-null because in the constructor, it is always set from the offline handler if an offline handler is present: https://github.com/Flagsmith/flagsmith-dotnet-client/blob/4327e266a67f721c2c104ad5df4298d49b220436/Flagsmith.FlagsmithClient/FlagsmithClient.cs#L130-L133

## Requested Changes
Please change the logic of the `FlagsmithClient` such that the `Environment` model is not automatically registered from the offline handler during construction if offline mode is false. I am not sure what-all considerations will need to be made in order to make this change, but that's the basic idea.

[...]

PR #110, opened 2024-07-23
fix: OfflineHandler causes the SDK to bypass live flag lookups

novakzaballa:
Fixing https://github.com/Flagsmith/flagsmith-dotnet-client/issues/103

- An additional validation was added to avoid always using the offlineHandler document when it is set.

- Added a new test for this specific case

### DIFF
--- a/Flagsmith.Client.Test/FlagsmithTest.cs
+++ b/Flagsmith.Client.Test/FlagsmithTest.cs
@@ -53,7 +53,7 @@ namespace Flagsmith.FlagsmithClientTest
             mockHttpClient.verifyHttpRequest(HttpMethod.Get, "/api/v1/environment-document/", Times.Once);
             await flagsmithClientTest.GetEnvironmentFlags();
             mockHttpClient.verifyHttpRequest(HttpMethod.Get, "/api/v1/environment-document/", Times.Once);
-            mockHttpClient.verifyHttpRequest(HttpMethod.Get, "/api/v1/flags/", Times.Once);
+            mockHttpClient.verifyHttpRequest(HttpMethod.Get, "/api/v1/flags/", Times.Never);
         }
         [Fact]
         public async Task TestGetEnvironmentFlagsCallsApiWhenNoLocalEnvironment()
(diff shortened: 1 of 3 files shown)


Flagsmith.Client.Test/FlagsmithTest.cs
```diff
@@ -391,6 +434,31 @@ public async Task TestFlagsmithUsesOfflineHandlerIfSetAndNoAPIResponse()
...
+                Content = new StringContent(Fixtures.ApiFlagResponse),
+            });
+
+            // When
+            FlagsmithClient flagsmithClientTest = new FlagsmithClient(Fixtures.ApiKey, httpClient: mockHttpClient.Object, offlineHandler: mockOfflineHandler.Object);
+
+            // Then
+            var environmentFlags = await flagsmithClientTest.GetEnvironmentFlags();
+            Assert.True(await environmentFlags.IsFeatureEnabled("some_feature"));
+            Assert.Equal("some-value", await environmentFlags.GetFeatureValue("some_feature"));
```
matthewelwell, line 459: Could we explicitly check that the `mockOfflineHandler` wasn't called here as well perhaps?

Flagsmith.FlagsmithClient/FlagsmithClient.cs
```diff
@@ -203,7 +203,7 @@ public async Task<IFlags> GetEnvironmentFlags()
 
         private async Task<IFlags> GetFeatureFlagsFromCorrectSource()
         {
-            return Environment != null ? GetFeatureFlagsFromDocument() : await GetFeatureFlagsFromApi().ConfigureAwait(false);
+            return OfflineMode && Environment != null ? GetFeatureFlagsFromDocument() : await GetFeatureFlagsFromApi().ConfigureAwait(false);
```
matthewelwell, line 206: Hmm... won't this break local evaluation mode as well? Shouldn't this be something like `(OfflineMode || LocalEvaluationEnabled) && ...`?

novakzaballa: You’re right, Done

Flagsmith.Client.Test/FlagsmithTest.cs
novakzaballa: Actually, it should only make a call based on this condition.
https://github.com/Flagsmith/flagsmith-dotnet-client/blob/0a2ba7950cee37f676d2b4db16f6f52884f83dae/Flagsmith.FlagsmithClient/FlagsmithClient.cs#L130

matthewelwell: Yes, but the test should verify that, right?

novakzaballa: I added 2 checks:

- Verify that the endpoint `api/v1/flags` is called once to retrieve its values.
- Check that the flag value is not the same as the value in the environment file.

matthewelwell (approved):

PR #115, opened 2024-09-04
deps: bump artifact options

matthewelwell:
Supersedes #114 which only updated download-artifact (which is stated as not being backwards compatible with artifacts uploaded with v3 [here](https://github.com/actions/download-artifact/releases/tag/v4.0.0)).


matthewelwell:
Note that I tested this PR by temporarily updating the release workflow to run through the steps that use the upload- and download-artifact actions and verified that it worked as expected.

zachaysan (approved):

Issue #113, opened 2024-09-03
Still experiencing hang in MVC View with enableAnalytics is true


DavidPerish:
In reference to #106 which was closed as duplicated of #105, it looks like we're still seeing the hang in MVC Views, specifically when `enableAnalytics` is set to true on the client.

Would it be possible to also add `ConfigureAwait(false)` to the `HttpClient.SendAsync` call in AnalyticsProcessor.cs? 

Thanks.

matthewelwell:
Hi @DavidPerish , thanks for raising this. That sounds sensible to me. Would you be willing to submit a PR for that?

DavidPerish:
@matthewelwell, forked, made the change, ran all tests and created the PR here:  https://github.com/Flagsmith/flagsmith-dotnet-client/pull/116

matthewelwell:
Awesome! Thanks @DavidPerish , we'll review the PR as soon as we can!

rolodato:
Merged and released in [5.4.0](https://github.com/Flagsmith/flagsmith-dotnet-client/releases/tag/v5.4.0). Thanks!

PR #116, opened 2024-09-05
Add ConfigureAwait(false) to AnalyticsProcessor

DavidPerish:
This adds ConfigureAwait(false) to AnalyticsProcessor.cs, per #113.


DavidPerish:
Had additional commit not being picked up

PR #130, opened 2025-01-14
fix: missing semver dependency

matthewelwell:
Fixes an issue using semver based segments in local evaluation mode. The following error would occur: 

```
Unhandled exception. System.IO.FileNotFoundException: Could not load file or assembly 'Semver, Version=2.1.0.0, Culture=neutral, PublicKeyToken=null'. The system cannot find the file specified.
```

This is seemingly caused by the fact that we're using `ProjectReference` to reference the Engine project which isn't including the subsequent `PackageReference`s in that project when building the FlagsmithClient project. 

A more long term solution is likely to `dotnet pack` the Flagsmith.Engine project and add it as a `PackageReference` in the `FlagsmithClient` project. 

Note that we also bump to the latest minor version as part of this PR.


matthewelwell:
So, my understanding of the issues I'm seeing in this PR is that something has changed in the Github runner which means that the version we are referencing in global.json no longer exists on the runners. You can see in the following screenshot that previous runs of this workflow have passed, but not as we expect - they are just running the tests against 7.0.43 instead of the version configured in the matrix. 

image

We probably have a few options to resolve this: 

 1. Get rid of matrix testing and just test on a single version (this is probably a bad idea)
 2. Remove global.json (although that is now inexplicably causing an issue with the EngineTest project, see [here](https://github.com/Flagsmith/flagsmith-dotnet-client/actions/runs/12773588085/job/35605838880))
 3. Maybe something else?

tiagoapolo (approved):

Issue #135, opened 2025-01-21
Deprecate FlagsmithClient constructors with more than 1 parameter


rolodato:
I want to rename `EnableClientSideEvaluation` to `EnableLocalEvaluation` for consistency with every other Flagsmith SDK. This is currently impossible without introducing breaking changes:

The [`FlagsmithClient` constructor has explicit ordered parameters](https://github.com/Flagsmith/flagsmith-dotnet-client/blob/4f760b982ded4559a768b8933a4388b9e5690341/Flagsmith.FlagsmithClient/FlagsmithClient.cs#L90). We can only add optional parameters at the end, and not rename any existing parameters to avoid breaking changes. This constructor is the [main one used in our docs](https://docs.flagsmith.com/clients/server-side?language=dotnet#configuring-the-sdk). Customers might be using named parameters, but nothing prevents them from using ordered parameters like we do [in our own constructor](https://github.com/Flagsmith/flagsmith-dotnet-client/blob/137590cc3fd8d9c2b739d046618631f07ee1068f/Flagsmith.FlagsmithClient/FlagsmithClient.cs#L176-L188) which exclusively uses ordered parameters.

We do have another [constructor with a dedicated configuration object](https://github.com/Flagsmith/flagsmith-dotnet-client/blob/137590cc3fd8d9c2b739d046618631f07ee1068f/Flagsmith.FlagsmithClient/FlagsmithClient.cs#L174) which also has a second `HttpClient` parameter, for some reason I don't understand, instead of being incorporated into the `FlagsmithConfiguration` object. Using this constructor is slightly less ergonomic than the previous one, but gives us more flexibility to manipulate the public API for initialising clients:

```csharp
new FlagsmithClient("ser.my-api-key", enableClientSideEvaluation: true, ...)
// vs
new FlagsmithClient(new FlagsmithConfiguration
{
    EnvironmentKey = "ser.my-api-key"
});
```

This constructor is currently only referenced in the "Singleton Initialisation" section on the same docs.

To avoid duplicating parameter documentation and having redundant constructors, we should introduce a new `Flagsmith(IFlagsmithConfiguration)` constructor, and deprecate all others. Later  we can add a Builder pattern to construct clients, which will be backwards compatible with the new single-argument constructor, and is slightly less verbose than manually instantiating a `FlagsmithConfiguration` object.

rolodato:
[...]

PR #136, opened 2025-01-21
feat: Deprecate constructors with >1 parameter and IFlagsmithConfig interface. Add EnableLocalEvaluation and EnvironmentRefreshInterval properties

rolodato:
Fixes https://github.com/Flagsmith/flagsmith-dotnet-client/issues/135.

Validated this does not break existing code by running the existing test suite without changes: https://github.com/Flagsmith/flagsmith-dotnet-client/pull/136/commits/0b41d2b7edcf26ac218f369e399335717637830f

### DIFF
--- a/Flagsmith.FlagsmithClient/FlagsmithClient.cs
+++ b/Flagsmith.FlagsmithClient/FlagsmithClient.cs
@@ -80,7 +80,7 @@ namespace Flagsmith
                         );
                     }
 
-                    _pollingManager = new PollingManager(GetAndUpdateEnvironmentFromApi, _config.EnvironmentRefreshIntervalSeconds);
+                    _pollingManager = new PollingManager(GetAndUpdateEnvironmentFromApi, _config.EnvironmentRefreshInterval);
                     Task.Run(async () => await _pollingManager.StartPoll()).GetAwaiter().GetResult();
                 }
             }
@@ -144,7 +144,7 @@ namespace Flagsmith
             {
                 EnvironmentKey = environmentKey,
                 ApiUri = new Uri(apiUrl),
-                EnvironmentRefreshIntervalSeconds = environmentRefreshIntervalSeconds,
+                EnvironmentRefreshInterval = TimeSpan.FromSeconds(environmentRefreshIntervalSeconds),
                 EnableLocalEvaluation = enableClientSideEvaluation,
                 Logger = logger,
                 EnableAnalytics = enableAnalytics,
--- a/Flagsmith.FlagsmithClient/FlagsmithClient.cs
+++ b/Flagsmith.FlagsmithClient/FlagsmithClient.cs
@@ -92,7 +92,7 @@ namespace Flagsmith
                 _flagListCacheDictionary = new ConcurrentDictionary<string, IdentityFlagListCache>();
             }
         }
-        
+
         public FlagsmithClient(FlagsmithConfiguration configuration)
         {
             _config = configuration;
--- a/Flagsmith.FlagsmithClient/FlagsmithClient.cs
+++ b/Flagsmith.FlagsmithClient/FlagsmithClient.cs
@@ -4,7 +4,6 @@ using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Linq;
-using System.Net;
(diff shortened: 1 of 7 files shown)


Flagsmith.FlagsmithClient/FlagsmithClient.cs
```diff
@@ -37,35 +36,75 @@ namespace Flagsmith
...
+                if (_config.EnableAnalytics)
+                    _analyticsProcessor = new AnalyticsProcessor(_config.HttpClient, _config.EnvironmentKey, _config.ApiUri.ToString(), _config.Logger, _config.CustomHeaders);
+
+                if (_config.EnableLocalEvaluation)
+                {
+                    if (!_config.EnvironmentKey!.StartsWith("ser."))
+                    {
+                        Console.WriteLine(
+                            "In order to use local evaluation, please generate a server key in the environment settings page."
+                        );
```
matthewelwell, line 80: This may conflict with #141, but as per the process laid out in this PR to _revert_ that change, I think we should keep this Console.WriteLine until we are ready to remove the deprecated methods in this PR.

Flagsmith.FlagsmithClient/FlagsmithConfiguration.cs
```diff
@@ -1,37 +1,72 @@
...
         }
+
         /// <summary>
-        /// Override the URL of the Flagsmith API to communicate with.
+        /// Versioned base Flagsmith API URI to use for all requests. Defaults to
+        /// <c>https://edge.api.flagsmith.com/api/v1/</c>.
+        /// <example><code>new Uri("https://flagsmith.example.com/api/v1/")</code></example>
         /// </summary>
-        public string ApiUrl { get; set; }
+        public Uri ApiUri { get; set; } = DefaultApiUri;
```
matthewelwell, line 30: Since the initial point of this PR was to bring consistency between the server side SDKs (at least that's my understanding), I'm surprised to see us change the name of this argument? Why not just keep using the string and build the URI ourselves?

matthewelwell (commented):
I've left a few comments but overall looks good. I'm not sure if this comment from the PR description is still valid though as it looks like there have been more changes to client / interface since then? 

> Validated this does not break existing code by running the existing test suite without changes: https://github.com/Flagsmith/flagsmith-dotnet-client/commit/0b41d2b7edcf26ac218f369e399335717637830f

Flagsmith.FlagsmithClient/FlagsmithConfiguration.cs
rolodato: Making server-side SDKs consistent in their parameter names was the initial driver for this PR. But, because the SDK constructor uses position-based and named arguments we can't remove or rename any arguments without introducing breaking changes.

By accepting the URI as an object and not a string, we don't have to deal with parsing URLs manually or error handling. FxCop has this as a default rule to enforce exactly this: https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1054.

My opinion is, if there's already a universally-used type for dealing with URIs, and it takes almost no effort to use, there's no reason to have to manipulate strings ourselves. And of course, this is exactly how our favourite competitor implemented this in their SDK.

rolodato: This is also the same reason why I deprecated `int EnvironmentRefreshIntervalSeconds` in favour of `TimeSpan EnvironmentRefreshInterval`.

rolodato:
> I've left a few comments but overall looks good. I'm not sure if this comment from the PR description is still valid though as it looks like there have been more changes to client / interface since then?

I added the previous tests as they're currently in `main` here: https://github.com/Flagsmith/flagsmith-dotnet-client/pull/136/commits/1753943b34349a6fdd201ec814a4f196d185c22e. We can delete this in the following major version.

Flagsmith.FlagsmithClient/FlagsmithConfiguration.cs
matthewelwell: Yep, ok, fair enough.

matthewelwell (approved):

PR #147, opened 2025-04-04
fix!: Remove deprecated methods and constructors. Throw error if using local eval without server-side key

rolodato:
Formatting changes introduced by `dotnet format`.


Flagsmith.FlagsmithClient/Flagsmith.FlagsmithClient.csproj
matthewelwell, line 1: I think your IDE wants to format XML as 4 spaces instead of 2 spaces? I'd prefer to keep this at 2 spaces unless there's a reason not to?

matthewelwell (commented):
Can we also re-add [this change](https://github.com/Flagsmith/flagsmith-dotnet-client/pull/141) to save dealing with merge conflicts?

Flagsmith.FlagsmithClient/Flagsmith.FlagsmithClient.csproj
matthewelwell: ok, then let's run with it 👍

matthewelwell (approved):

PR #175, opened 2025-11-18
fix(Engine): Fix segment condition evaluation across all system locales

emyller:
Fix segment condition evaluation across all system locales.

## Changes
- Apply locale-independent parsing to numeric strings in segment conditions

## How to reproduce

```bash
LANG=de_DE.UTF-8 LC_ALL=de_DE.UTF-8 dotnet test
```

## Test failure example

```
Failed: EngineTest.Unit.Segments.SegmentEvaluatorTest.TestSegmentConditionMatchesTraitValue(
  _operator: "EQUAL", 
  traitValue: 1,23, 
  conditionValue: "1.23", 
  expectedResult: True
)
Expected: True
Actual:   False
```

> [!NOTE]
> CI uses C.UTF-8 locale (period decimal separator). This bug could affect German, French, Spanish, Italian, Portuguese, and other locales using comma as decimal separator.

Review effort: 1/5

### DIFF
--- a/Flagsmith.Engine/Segment/Evaluator.cs
+++ b/Flagsmith.Engine/Segment/Evaluator.cs
@@ -8,7 +8,6 @@ using Newtonsoft.Json.Linq;
 using Semver;
 using System;
 using System.Collections.Generic;
-using System.Globalization;
 using System.Linq;
 
 namespace FlagsmithEngine.Segment
@@ -374,7 +373,7 @@ namespace FlagsmithEngine.Segment
             long conditionValue;
             try
             {
-                conditionValue = Convert.ToInt64(condition.Value, CultureInfo.InvariantCulture);
+                conditionValue = InvariantConvert.ToInt64(condition.Value);
             }
             catch (FormatException)
             {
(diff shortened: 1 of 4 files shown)


Flagsmith.Engine/Segment/Evaluator.cs
```diff
@@ -396,12 +397,12 @@ static bool intOperations(long traitValue, SegmentConditionModel condition)
         {
             switch (condition.Operator)
             {
-                case Constants.Equal: return traitValue == Convert.ToInt32(condition.Value);
-                case Constants.NotEqual: return traitValue != Convert.ToInt32(condition.Value);
-                case Constants.GreaterThan: return traitValue > Convert.ToInt32(condition.Value);
-                case Constants.GreaterThanInclusive: return traitValue >= Convert.ToInt32(condition.Value);
-                case Constants.LessThan: return traitValue < Convert.ToInt32(condition.Value);
-                case Constants.LessThanInclusive: return traitValue <= Convert.ToInt32(condition.Value);
+                case Constants.Equal: return traitValue == Convert.ToInt32(condition.Value, CultureInfo.InvariantCulture);
```
gagantrivedi: Should we consider making more dry? Maybe something like this:
```
 // Flagsmith.Engine/Utils/InvariantConvert.cs
  using System;
  using System.Globalization;

  namespace FlagsmithEngine.Utils
  {
      internal static class InvariantConvert
      {
          public static int ToInt32(string value) =>
              Convert.ToInt32(value, CultureInfo.InvariantCulture);

          public static long ToInt64(string value) =>
              Convert.ToInt64(value, CultureInfo.InvariantCulture);

          public static float ToSingle(string value) =>
              float.Parse(value, CultureInfo.InvariantCulture);

          public static double ToDouble(string value) =>
              Convert.ToDouble(value, CultureInfo.InvariantCulture);
      }
  }

  Then replace all conversions:
  // Before (current PR)
  traitValue == Convert.ToInt32(condition.Value, CultureInfo.InvariantCulture)

  // After (deeper abstraction)
  traitValue == InvariantConvert.ToInt32(condition.Value)
```

gagantrivedi (dismissed):

Flagsmith.Engine/Segment/Evaluator.cs
emyller: Addressed in 7d5f322.

khvn26 (approved):
👍

Issue #168, opened 2025-10-01
Integrate Context Value supported Engine into the SDK


matthewelwell:
This issue covers the work to integrate the SDK with the updated engine (as part of #166) which adds support for context values.

PR #177, opened 2025-11-18
feat(Client): Add mapper layer for context-based engine

emyller:
Add mapper layer to transform environment documents into evaluation contexts. No SDK integration in this PR.

This is pass 2 of 3.

1. Implement context-based engine (PR #171)
2. Add mapper layer. **No SDK changes.** (this PR)
3. SDK integration and legacy engine removal. **Breaking changes.** (future PR)

Mapper implementation inspired by:
- [flagsmith-python-client](https://github.com/Flagsmith/flagsmith-python-client/blob/main/flagsmith/mappers.py)
- [flagsmith-php-client#108](https://github.com/Flagsmith/flagsmith-php-client/pull/108)
- [flagsmith-php-client#109](https://github.com/Flagsmith/flagsmith-php-client/pull/109)
- [flagsmith-go-client#178](https://github.com/Flagsmith/flagsmith-go-client/pull/178)
- [flagsmith-nodejs-client#203](https://github.com/Flagsmith/flagsmith-nodejs-client/pull/203)

## Changes
- [x] Transform EnvironmentModel to EvaluationContext

Contributes to #168

Review effort: 4/5

### DIFF
--- a/Flagsmith.Client.Test/MappersTest.cs
+++ b/Flagsmith.Client.Test/MappersTest.cs
@@ -1,0 +1,132 @@
+using System.Linq;
+using FlagsmithEngine;
+using FlagsmithEngine.Segment;
+using Xunit;
+
+namespace Flagsmith.FlagsmithClientTest
+{
+    public class MappersTest
+    {
+        [Fact]
+        public void MapEnvironmentDocumentToContext_ProducesEvaluationContext()
+        {
+            // Given
+            var environment = Fixtures.Environment;
+
+            // When
+            var context = Mappers.MapEnvironmentDocumentToContext(environment);
+
+            // Then
+            Assert.IsType<EvaluationContext<SegmentMetadata, FeatureMetadata>>(context);
+            Assert.Equal("test_key", context.Environment.Key);
+            Assert.Equal("Test project", context.Environment.Name);
+            Assert.Null(context.Identity);
+            Assert.Equal(2, context.Segments.Count);
+
+            // Verify API segment
+            Assert.True(context.Segments.ContainsKey("1"));
+            var apiSegment = context.Segments["1"];
+            Assert.Equal("1", apiSegment.Key);
+            Assert.Equal("Test segment", apiSegment.Name);
+            Assert.Single(apiSegment.Rules);
+            Assert.Empty(apiSegment.Overrides);
+            Assert.Equal("api", apiSegment.Metadata.Source);
+            Assert.Equal(1, apiSegment.Metadata.Id);
+
+            // Verify segment rule structure
+            Assert.Equal(TypeEnum.All, apiSegment.Rules[0].Type);
+            Assert.Empty(apiSegment.Rules[0].Conditions);
+            Assert.Single(apiSegment.Rules[0].Rules);
+
+            Assert.Equal(TypeEnum.All, apiSegment.Rules[0].Rules[0].Type);
+            Assert.Single(apiSegment.Rules[0].Rules[0].Conditions);
+            Assert.Empty(apiSegment.Rules[0].Rules[0].Rules);
+
+            Assert.Equal("foo", apiSegment.Rules[0].Rules[0].Conditions[0].Property);
+            Assert.Equal(Operator.Equal, apiSegment.Rules[0].Rules[0].Conditions[0].Operator);
+            Assert.Equal("bar", apiSegment.Rules[0].Rules[0].Conditions[0].Value.String);
+
+            // Verify identity override segment
+            var overrideKey = "42d7556943d3c6f62b310e40f2252ac29203c20f37e9adffd8f12bd084a87b9d";
+            Assert.True(context.Segments.ContainsKey(overrideKey));
+            var overrideSegment = context.Segments[overrideKey];
+            Assert.Equal("", overrideSegment.Key);
+            Assert.Equal("identity_overrides", overrideSegment.Name);
+            Assert.Single(overrideSegment.Rules);
+            Assert.Single(overrideSegment.Overrides);
+
+            Assert.Equal(TypeEnum.All, overrideSegment.Rules[0].Type);
+            Assert.Single(overrideSegment.Rules[0].Conditions);
+            Assert.Empty(overrideSegment.Rules[0].Rules);
+
+            Assert.Equal("$.identity.identifier", overrideSegment.Rules[0].Conditions[0].Property);
+            Assert.Equal(Operator.In, overrideSegment.Rules[0].Conditions[0].Operator);
+            Assert.Equal(new[] { "overridden-id" }, overrideSegment.Rules[0].Conditions[0].Value.StringArray);
+
+            Assert.Equal("", overrideSegment.Overrides[0].Key);
+            Assert.Equal("some_feature", overrideSegment.Overrides[0].Name);
+            Assert.False(overrideSegment.Overrides[0].Enabled);
+            Assert.Equal("some-overridden-value", overrideSegment.Overrides[0].Value);
+            Assert.Equal(Constants.StrongestPriority, overrideSegment.Overrides[0].Priority);
+            Assert.Null(overrideSegment.Overrides[0].Variants);
+            Assert.Equal(1, overrideSegment.Overrides[0].Metadata.Id);
+
+            // Verify features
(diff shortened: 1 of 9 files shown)


Flagsmith.Client.Test/MappersTest.cs
```diff
@@ -0,0 +1,132 @@
...
+            // Given
+            var environment = Fixtures.Environment;
+
+            // When
+            var context = Mappers.MapEnvironmentDocumentToContext(environment);
+
+            // Then
+            Assert.IsType<EvaluationContext<SegmentMetadata, FeatureMetadata>>(context);
+            Assert.Equal("test_key", context.Environment.Key);
+            Assert.Equal("Test project", context.Environment.Name);
```
Zaimwa9: That's the project name no ? I believe we need to update the fixture

Flagsmith.FlagsmithClient/Mappers.cs
```diff
@@ -0,0 +1,294 @@
...
+        /// </summary>
+        public static EvaluationContext<SegmentMetadata, FeatureMetadata> MapEnvironmentDocumentToContext(
+            EnvironmentModel environmentDocument)
+        {
+            var context = new EvaluationContext<SegmentMetadata, FeatureMetadata>
+            {
+                Environment = new EnvironmentContext
+                {
+                    Key = environmentDocument.ApiKey,
+                    Name = environmentDocument.Project.Name,
```
Zaimwa9: Indeed, the fixture is misleading and was missing the name (same happened in [Node](https://github.com/Flagsmith/flagsmith-nodejs-client/pull/226))

Flagsmith.FlagsmithClient/Mappers.cs
```diff
@@ -0,0 +1,294 @@
...
+        {
+            var identity = new IdentityContext
+            {
+                Identifier = identifier,
+                Traits = traits?.ToDictionary(t => t.GetTraitKey(), t => t.GetTraitValue()) ?? new Dictionary<string, object>(),
+            };
+
+            context = context.Clone();
+            context.Identity = identity;
+            return context;
```
Zaimwa9, line 98: That's where we need to set `identityContext.key = djangoId` if exists

Zaimwa9: Setting the key `environment_identifier` will happen in a following PR ? (`get_enriched_context` in some implementations)

Flagsmith.FlagsmithClient/Mappers.cs
```diff
@@ -0,0 +1,294 @@
...
+            var segments = new Dictionary<string, SegmentContext<SegmentMetadata, FeatureMetadata>>();
+
+            foreach (var kvp in featuresToIdentifiers)
+            {
+                var serializedOverridesKey = kvp.Key;
+                var identifiers = kvp.Value;
+
+                var segment = new SegmentContext<SegmentMetadata, FeatureMetadata>
+                {
+                    Key = UnusedKey,
```
Zaimwa9, line 212: Pointing out the discrepancy between SDKs that for some we use 
`const segmentKey = `identity_override_${overrideHash}`

Zaimwa9 (changes requested):
Couple of NITs, one is blocking (`environment.name`). Thanks for the exhaustive test, unfortunately i'm unable to provide code improvements

Flagsmith.Client.Test/MappersTest.cs
emyller: Yes, great catch. Addressed in 6beff14e2f23953a5525eff673661ffd4aa41c45.

Flagsmith.FlagsmithClient/Mappers.cs
emyller: Fixed in 6beff14e2f23953a5525eff673661ffd4aa41c45.

Flagsmith.FlagsmithClient/Mappers.cs
emyller: > That's where we need to set `identityContext.key = djangoId` if exists

I believe we should not do that. The suggestion is solid, but it leads to inconsistent engine results. I've collected evidence in the following issues:
- https://github.com/Flagsmith/flagsmith-nodejs-client/issues/227
- https://github.com/Flagsmith/flagsmith-ruby-client/issues/92


> Setting the key `environment_identifier` will happen in a following PR ? (`get_enriched_context` in some implementations)

We had it implemented by @khvn26 in [PR 1/3](https://github.com/Flagsmith/flagsmith-dotnet-client/pull/171): https://github.com/Flagsmith/flagsmith-dotnet-client/blob/6beff14e2f23953a5525eff673661ffd4aa41c45/Flagsmith.Engine/Engine.cs#L41

Flagsmith.FlagsmithClient/Mappers.cs
emyller: True, I can confirm the discrepancy. I can also confirm the `key` of a virtual segment created out of identity overrides has no utility, so it could really have any value in it, or none. I see no changes needed here — let me know if you agree.

Flagsmith.Client.Test/MappersTest.cs
```diff
@@ -0,0 +1,132 @@
...
+            Assert.Equal(50.0, mvFeatureWithoutIds.Variants[0].Weight);
+            Assert.Equal(1, mvFeatureWithoutIds.Variants[0].Priority); // Second in sorted UUID order
+
+            Assert.Equal("option_x", mvFeatureWithoutIds.Variants[1].Value);
+            Assert.Equal(25.0, mvFeatureWithoutIds.Variants[1].Weight);
+            Assert.Equal(0, mvFeatureWithoutIds.Variants[1].Priority); // First in sorted UUID order
+
+            Assert.Equal("option_z", mvFeatureWithoutIds.Variants[2].Value);
+            Assert.Equal(25.0, mvFeatureWithoutIds.Variants[2].Weight);
+            Assert.Equal(2, mvFeatureWithoutIds.Variants[2].Priority); // Third in sorted UUID order
```
khvn26: nit: Consider converting this to a single `Assert.Equivalent` assertion. This would allow us to have an idea of what `expectedContext` should look like, and to catch unexpected values.

Flagsmith.Client.Test/MappersTest.cs
```diff
@@ -0,0 +1,132 @@
...
+            var mvFeatureWithoutIds = context.Features["mv_feature_without_ids"];
+            Assert.Equal("3", mvFeatureWithoutIds.Key);
+            Assert.Equal("mv_feature_without_ids", mvFeatureWithoutIds.Name);
+            Assert.False(mvFeatureWithoutIds.Enabled);
+            Assert.Equal("fallback_value", mvFeatureWithoutIds.Value);
+            Assert.Null(mvFeatureWithoutIds.Priority);
+            Assert.Equal(3, mvFeatureWithoutIds.Variants.Length);
+            Assert.Equal(3, mvFeatureWithoutIds.Metadata.Id);
+
+            // Variants should be ordered by UUID alphabetically
```
khvn26: Order does not matter now, given we have the `priority` field — another argument in favour of using `Assert.Equivalent`.

khvn26 (dismissed):
LGTM overall, with optional test-related nitpicks 👍

Flagsmith.Client.Test/MappersTest.cs
emyller: I did spend some time now on this, and it worked. However, I realised I've actually 80% worked on this issue: https://github.com/Flagsmith/flagsmith-dotnet-client/issues/180
Do you mind if we push code to a new PR instead?

Flagsmith.Client.Test/MappersTest.cs
emyller: See [other comment](https://github.com/Flagsmith/flagsmith-dotnet-client/pull/177#discussion_r2560323988).

Zaimwa9 (approved):
👍

Flagsmith.Client.Test/MappersTest.cs
khvn26: Works for me 👍

khvn26 (approved):

PR #179, opened 2025-11-24
feat(Client)!: Integrate SDK with context-based engine

### LINKED ISSUE #168: Integrate Context Value supported Engine into the SDK

emyller:
Support context values in local evaluation.

This is pass 3 of 3.

1. Implement context-based engine (PR #171)
2. Add mapper layer. **No SDK changes.** (PR #177)
3. SDK integration and legacy engine removal. **Breaking changes.** (this PR)

## Changes
- [x] Evaluate flags using context-based engine.
- [x] Update engine test data to v3.4.2.
- [x] Clean up all unused code from old engine — still need Environment model and friends.

Closes #168

Review effort: 3/5 (the amount of deleted lines barks, but doesn't bite to 4/5)


khvn26 (changes requested):
I expect this PR to remove the entire `FlagsmithEngine.Models` namespace as well.

emyller:
> I expect this PR to remove the entire `FlagsmithEngine.Models` namespace as well.

I've went ahead and deleted the most code _that could only be deleted after the engine rewrite_ that I could find. Commit a060ce357d759290fb666e568da872b85e6ce6f9 may be easier to review isolated.

Flagsmith.Engine/Segment/Evaluator.cs
```diff
@@ -276,45 +273,6 @@ private static FlagResult<FeatureMetadataT> GetFlagResult<_, FeatureMetadataT>(E
...
-            }
-            else if (condition.Operator == Constants.IsNotSet)
-            {
-                return trait == null;
-            }
-
-            return trait != null && MatchesTraitValue(trait.TraitValue, condition);
-        }
-
         public static bool MatchesTraitValue(object traitValue, SegmentConditionModel condition)
```
khvn26: I expect this function to change to:

```suggestion
        public static bool MatchesContextValue(object contextValue, Condition condition)
```

Flagsmith.FlagsmithClient/FlagsmithClient.cs
```diff
@@ -162,15 +163,24 @@ public async Task<IFlags> GetIdentityFlagsFromCorrectSource(IdentityWrapper iden
...
+                _evaluationContext,
+                identifier,
+                traits);
 
-            return segmentModels?.Select(t => new Segment(id: t.Id, name: t.Name)).ToList<ISegment>();
+            var result = _engine.GetEvaluationResult(context);
+
+            var segments = result.Segments
+                .Where(s => s.Metadata.Id != null) // Not a real segment, e.g. an identity override virtual segment
+                .Select(s => new Segment(id: s.Metadata.Id.Value, name: s.Name))
```
khvn26: nit: Interesting how dotnet doesn't infer that `s.Metadata.Id` is not null at this point. Maybe there's a way to avoid this warning without silencing it; otherwise, I'd settle on an ignore comment here.

khvn26 (changes requested):
Very good, just a couple comments now.

Flagsmith.Engine/Segment/Evaluator.cs
emyller: Good catch. Fixed in 5a0fe23f4c41461acb5620f7230b5707083227c5.

khvn26: Not quite — it still accepts the old engine `SegmentConditionModel` class. I expect it to accept `Condition`.

Flagsmith.FlagsmithClient/FlagsmithClient.cs
emyller: Such warnings have gone unnoticed by me so far — took me a while to make sense of this comment. Thanks for pointing me towards it.

Fixed this warning:
```
Flagsmith.FlagsmithClient/FlagsmithClient.cs(180,46): warning CS8629: Nullable value type may be null.
```

And also this one introduced recently in #177:
```
Flagsmith.FlagsmithClient/Metadata.cs(27,22): warning CS8632: The annotation for nullable reference types should only be used in code within a '#nullable' annotations context.
```

See 31caa660f952d62ee9ff3d83ac04115cf15bb445.

Flagsmith.Engine/Segment/Evaluator.cs
emyller: Attempted with 1734178856088312e1d14a85796ed7f2d345daa7.

khvn26 (dismissed):
LGTM 👍

matthewelwell (dismissed):

khvn26 (dismissed):
Overall, the new additions (or, rather, removals) look great to me.

The only comment I have is maybe namespace the old models to indicate they're only being used by the API — I did it in the Python SDK by consolidating them in the `flagsmith.api.types` module.

emyller:
> The only comment I have is maybe namespace the old models to indicate they're only being used by the API — I did it in the Python SDK by consolidating them in the `flagsmith.api.types` module.

I've considered something similar (moving models to the `FlagsmithClient` package), and spent some time discussing with Claude, and concluded that I'd have no benefit with adding review effort refactoring code we mean to delete in a later stage. I used the strategy of drying out old engine code enough to allow for an easy `rm` (along with mappers) whenever the `EnvironmentModel` is no longer needed. That said, I don't oppose to the refactor path if you prefer and don't mind the review effort. I kind of have it half sketched, but decided to leave it off if you agree.

Flagsmith.Engine/Segment/SegmentConditionOperator.cs
khvn26, line 1: nit: I understand the preference to have a single operator enum. However, I would err on leaving the enum from `EvaluationContext` in since a) strictly, the engine schema should be the source of truth and b) that code is auto-generated from said schema.

Given the auto-generated classes still had to be modified and adopting the enum would likely be more painful, this is a non-blocking comment.

khvn26:
> That said, I don't oppose to the refactor path if you prefer and don't mind the review effort. I kind of have it half sketched, but decided to leave it off if you agree.

I agree. I don't think it's worth to hold off releasing the SDK further for this.

Flagsmith.Engine/Segment/SegmentConditionOperator.cs
emyller: Do we auto-generate code I just modified?
image

I can revert the original type back if that's the right thing to do here. So far being naive of this auto-generation machanism, the `enum` seemed like an idiomatic option carrying adaptation burden.

emyller: I've necromanced it back with d70373c58ededcec1ae0dc78a472b4e0b35b48f9.

khvn26: Nice 👍

khvn26 (approved):

PR #181, opened 2025-11-27
fix: Exclude identities when PERCENTAGE_SPLIT trait is undefined

### LINKED ISSUE #178: Exclude identities when PERCENTAGE_SPLIT trait is undefined
Users without specified trait appear in segment, evaluated using identity key instead.

## Acceptance criteria
- [ ] Return `false` when trait is undefined instead of falling back to identity key

**Current implementation:** https://github.com/Flagsmith/flagsmith-dotnet-client/blob/3a25c0228c10cbbbd28f01f8b77ffcc0a7e2e031/Flagsmith.Engine/Segment/Evaluator.cs#L174-L175

See https://github.com/Flagsmith/flagsmith-engine/issues/283 for context.

khvn26:
Closes #178.

### DIFF
--- a/Flagsmith.EngineTest/EngineTest.cs
+++ b/Flagsmith.EngineTest/EngineTest.cs
@@ -49,7 +49,7 @@ namespace EngineTest
             var result = _iengine.GetEvaluationResult(testCase.Context);
 
             // Then
-            Assert.Equivalent(testCase.Result, result);
+            Assert.Equivalent(testCase.Result, result, strict: true);
         }
 
         public static IEnumerable<object[]> ExtractTestCaseFilenames()
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,4 +1,4 @@
 [submodule "Flagsmith.EngineTest/EngineTestData"]
 	path = Flagsmith.EngineTest/EngineTestData
 	url = git@github.com:Flagsmith/engine-test-data.git
-	tag = v3.4.2
+	tag = v3.5.0
--- a/Flagsmith.Engine/Engine.cs
+++ b/Flagsmith.Engine/Engine.cs
@@ -203,7 +203,7 @@ namespace FlagsmithEngine
 
                     if (contextValue != null)
                         objectIds = new List<string> { segmentKey, contextValue.ToString() };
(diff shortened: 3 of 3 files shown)


Flagsmith.EngineTest/EngineTest.cs
```diff
@@ -49,7 +49,7 @@ public void Test_Engine(String testCaseFilename)
             var result = _iengine.GetEvaluationResult(testCase.Context);
 
             // Then
-            Assert.Equivalent(testCase.Result, result);
+            Assert.Equivalent(testCase.Result, result, strict: true);
```
khvn26, line 52: Confusingly, this didn't cause any discrepancies for other tests, but had to be added for `test_percentage_split__missing_trait__should_not_match` produce the correct assertion.

emyller: It's amazing how difficult it is to find manuals for this. These [release notes](https://xunit.net/releases/v2/2.4.2#new-assertions) are the only source of education I could find. Nothing to act here if it works, but this post scriptum from release notes makes me a bit uncomfortable:

> It's somewhat challenging to describe these rules in plain English, so please see the comprehensive test suite linked below for codified versions of all these rules.
[xunit/xunit#1604](https://github.com/xunit/xunit/issues/1604)
[Assert.Equivalent test suite](https://github.com/xunit/xunit/blob/f110e5bee5dfd4c08339587c9c3df9292fcb597c/test/test.xunit.assert/Asserts/EquivalenceAssertsTests.cs)

Also worth noting v3 offers easier to read "[Remarks](https://api.xunit.net/v3/3.0.1/v3.3.0.1-Xunit.Assert.Equivalent.html#v3_3_0_1_Xunit_Assert_Equivalent_System_Object_System_Object_System_Boolean__remarks)", which makes it unclear if the functionality is the same compared to v2.

emyller (approved):
LGTM

Flagsmith.EngineTest/EngineTest.cs
khvn26: Judging from the linked test suite, `strict: true` behaviour is what we need for our tests.

